PR #2, opened 2018-01-04
[#3] Add Rat type and typefamilies, add Units

vrom911:


src/Time/Rational.hs
```diff
@@ -0,0 +1,32 @@
...
+       , Gcd
+       , Normalize
+       ) where
+
+import GHC.TypeNats
+
+-- | Data structure represents the rational number.
+data Rat = MkRat Nat Nat
+
+type (%) = 'MkRat
```
int-index: I thought about it a bit more.

A better name for `MkRat` is `:%`, and `%` should also perform normalization:

```
data Rat = Nat :% Nat

type family (a :: Nat) % (b :: Nat) :: Rat where
  a % b = Normalize (a :% b)
```

src/Time/Rational.hs
```diff
@@ -0,0 +1,32 @@
...
+infixl 7 %
+
+type family DivRat (a :: Rat) (b :: Rat) :: Rat where
```
ChShersh: Probably better to replace `a` and `b` with `m` and `n`. To avoid name collisions with definition of `DivRat`. I think it's better to use unique names in type signature. Also it makes it more consisten with naming in standard library:

* https://hackage.haskell.org/package/base-4.10.1.0/docs/GHC-TypeNats.html#t:-43-

src/Time/Rational.hs
```diff
@@ -0,0 +1,32 @@
...
+
+module Time.Rational
+       ( Rat
+       , type (%)
+       , DivRat
+       , Gcd
+       , Normalize
+       ) where
+
+import GHC.TypeNats
```
ChShersh: Explicit imports :slightly_smiling_face:

src/Time/Rational.hs
```diff
@@ -0,0 +1,32 @@
...
+-- | Data structure represents the rational number.
+data Rat = MkRat Nat Nat
+
+type (%) = 'MkRat
+infixl 7 %
+
+type family DivRat (a :: Rat) (b :: Rat) :: Rat where
+    DivRat (a % b) (c % d) = Normalize ((a * d) % (b * c))
+
+type family Gcd (a :: Nat) (b :: Nat) :: Nat where
```
ChShersh: I think it's a good idea to add a couple Haddock examples to every type family with `:kind!`. Also, we probably can write some tests for these type families later on...

src/Time/Units.hs
```diff
@@ -0,0 +1,64 @@
...
+
+type SecondUnit      = 1 % 1
+type MilliSecondUnit = 1 % 1000
+type MicroSecondUnit = 1 % 1000000
+type NanoSecondUnit  = 1 % 1000000000
+
+type MinuteUnit      = 60 % 1
+type HourUnit        = 360 % 1
+type DayUnit         = 8640 % 1
+type WeekUnit        = 60480 % 1
```
ChShersh: I think it's a good idea to add a couple extra families which allow us to specify this time units in more convenient way. Something like this:

```haskell
type WeekUnit = (5 % 1) %* Day
```

This can be achieved with the following type family:

```haskell
type family (%*) (m :: Rat) (n :: Rat) where ...
```

Not sure about `%*` name, but `*` is already taken for `Nat`...
  
Maybe for even nicer API we might want to write something like this:

```
type MilliSecondUnit = Second / 1000
type MinuteUnit = 60 :* Second
```

Which results in couple extra couple type families. What do you think?

src/Time/Units.hs
```diff
@@ -0,0 +1,64 @@
...
+       , WeekUnit
+       , FortnightUnit
+       ) where
+
+import Data.Ratio (Rational)
+
+import Time.Rational (type (%), Rat)
+
+
+newtype Time (rat :: Rat) = Time Rational
```
ChShersh: You can derive `Num` instance for this data type using `GeneralizedNewtypeDeriving` extension. This instance would be useful.

Also, I guess, you need to replace `Rational` with `Ratio Natural`. `Rational` is `Ratio Integer`. We use `GHC.TypeNats` module which allows to reflect `Nat` from type into `Natural` data type, not `Integer`. 

I asked a question about difference between `Natural` and `Integer` some time ago:

* https://stackoverflow.com/questions/45904898/when-to-pick-natural-over-integer-in-haskell

int-index: Is negative time impossible in your model?

ChShersh: @int-index I didn't find reasons and use-cases for having negative time.

int-index: What if I use the `time` type to represent a timestamp starting from 0 AD and then I have a BC date?

ChShersh: @int-index Are you aware of `Either` data type?

int-index: Furthermore, I think there should be two types:

1. Timestamp — can be negative, has no units
2. Time (TimeDiff) — cannot be negative, has units

and instead of `Num` instances that provides subtraction, there must be operations like:

```
data DiffSign = Before | After

timediff :: KnownRat unit => Timestamp -> Timestamp -> (DiffSign, Time unit)
timeadd :: KnownRat unit => Time unit -> Time unit -> Time unit
timemult :: KnownRat unit => Ratio Natural -> Time unit -> Time unit
```

int-index: >  Are you aware of Either data type?

I know I can add a sign externally, but this would mean I have to redefine arithmetic operations. Anyway, now that I think of it, arithmetic operations on time don't fit into the `Num` hierarchy anyway (see the comment above).

int-index: To make my reasoning more intuitive, imagine a timeline. `Timestamp` is a point on this line, and `Time unit` is an interval measured in `unit`.

ChShersh: @int-index What `Timestamp` should represent? Is it microseconds or nanoseconds? If `Timestamp` is `Integer` for example? What `Timestamp` stored internally?

int-index: > What Timestamp should represent?

A point on the timeline.

> Is it microseconds or nanoseconds?

It's unitless.

> If Timestamp is Integer for example? What Timestamp stored internally?

Arbitrary precision, it's `Rational`.

vrom911:
image
Here is how piece of haddock looks like now

src/Time/Rational.hs
```diff
@@ -1,8 +1,14 @@
...
 {-# LANGUAGE DataKinds            #-}
 {-# LANGUAGE ExplicitNamespaces   #-}
+{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE GADTs                #-}
 {-# LANGUAGE KindSignatures       #-}
+{-# LANGUAGE Rank2Types           #-}
+{-# LANGUAGE ScopedTypeVariables  #-}
 {-# LANGUAGE TypeFamilies         #-}
 {-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE TypeSynonymInstances #-}
```
ChShersh: Probably `AllowAmbiguousTypes`, `GADTs`, `Rank2Types` are not needed... I'm not sure but you probably can clean up extensions.

src/Time/Rational.hs
```diff
@@ -114,3 +127,17 @@ Normalize (9 % 12) :: Rat
...
+
+-- | Rational numbers, with numerator and denominator of 'Natural' type.
+type RatioNat = Ratio Natural
+
+-- | This class gives the integer associated with a type-level rational.
+class KnownRat (r :: Rat) where
+  ratVal :: Proxy r -> Ratio Natural
+
+instance (KnownNat a, KnownNat b) => KnownRat (a :% b) where
+    ratVal _ = natVal (Proxy :: Proxy a) :% natVal (Proxy :: Proxy b)
```
ChShersh: If you add `TypeApplications` extension you can write this as `Proxy @a` and `Proxy @b`.

src/Time/Units.hs
```diff
@@ -36,15 +36,18 @@ module Time.Units
         -- ** Functions
        , convertUnit
        , intPart
+       , threadDelay
        ) where
 
+import Control.Monad.IO.Class (MonadIO, liftIO)
```
ChShersh: You need to add `transformers` package to dependencies in order to make this work

* http://hackage.haskell.org/package/transformers

I guess `cabal` uses some global cache. That's why it didn't warn you that it doesn't know `Control.Monad.IO.Class` module. Whether we want to add dependency on `transformers` package or not is another question... But, I guess, there's not harm. `MonadIO` version is much useful as for me.

src/Time/Units.hs
```diff
@@ -85,3 +88,11 @@ convertUnit (Time ratNat) = Time $ ratNat * (ratVal $ divRat (Proxy @unitFrom) (
 -- | Extracts integer part of time unit.
 intPart :: Time unit -> Natural
 intPart (Time n) = numerator n `div` denominator n
+
+-- | Convenient version of 'Control.Concurrent.threadDelay' which takes
+-- any time-unit and operates in any MonadIO.
```
ChShersh: `threadDelay` from `Control.Concurrent` takes `Int` while you're converting `Natural` to `Int`. `Natural` is arbitrary precision number. `Int` has upper bound in `9 * 10^18` approximately. Maybe we should add warning in documentation that people should pass bigger time units? :slightly_smiling_face: Well, in reality people won't pass such big numbers. But they accidentally can... And we probably shouldn't fail with exception. When you implement playground, I want you to test behavior with big time units, how big `Natural` converts to `Int` and also (if required) check how `threadDelay` behaves when you pass negative `Int` to it.

src/Time/Units.hs
```diff
@@ -85,3 +88,11 @@ convertUnit (Time ratNat) = Time $ ratNat * (ratVal $ divRat (Proxy @unitFrom) (
...
+threadDelay :: forall unit m .
+               (KnownRat unit, KnownRat (DivRat unit MicroSecondUnit), MonadIO m)
```
ChShersh: We need to do something to not specify `KnownRat` for `DivRat`... I don't know how to do it, need to think more.

src/Time/Units.hs
vrom911: I figured out that very big `Natural` gives negative number when using convertion to `Int`. But function `threadDelay` from `Concurrent` works without exception on that cases, without any delay. My example was
```
let billionDays = Time 1000000000 :: Day
    let mcs :: Int = fromIntegral $ intPart $ (convertUnit billionDays :: MicroSecond)
    putStrLn $ "Billion days is " ++ show mcs ++ " microseconds"
    Concurrent.threadDelay (-5000000)
    putStrLn "Hmm"
```
and immediate result
```
Billion days is -5833720368547758080 microseconds
Hmm
```

test/Test/Time/TypeSpec.hs
```diff
@@ -0,0 +1,103 @@
...
+     -*- It "GCD 9 1000 = 1" (Gcd 9 1000 `Is` 1)
+     -*- It "GCD 1000 9 = 1" (Gcd 1000 9 `Is` 1)
+
+ -/-
+
+    "Common divisor"
+    ~~~~~~~~~~~~
+         It "GCD 2 6 = 2"      (Gcd 2 6      `Is` 2)
+     -*- It "GCD 3 6 = 3"      (Gcd 3 6      `Is` 3)
+     -*- It "GCD 500 1000 = 1" (Gcd 500 1000 `Is` 500)
```
vrom911: Small typo `"GCD 500 1000 = 500"`

src/Time/Units.hs
ChShersh: @vrom911 Thanks for your investigation! I wasn't aware of such behavior. Well, probably we shouldn't add extra warning to our `threadDelay` function... Users have the same chances to fuck up with `threadDelay` from `Control.Concurrent` module. I guess we didn't introduce any new bugs. Even more, our `threadDelay` is safer. Because unexpected behavior for `threadDelay` can be achieved by passing negative `Int` while with `o-clock` you can't pass negative `Time`. You will get negative time only in case of **really** big time.

src/Time/Units.hs
int-index: This sort of redundant constraints is solvable via typechecker plugins such as https://github.com/clash-lang/ghc-typelits-natnormalise, we might need our own for `KnownRat`.

Alternatively, we can use `Dict` and `unsafeCoerce` to manipulate dictionaries manually and produce desired `KnownRat` constraints, i.e.

```
withDivRat :: (KnownRat a, KnownRat b) => (KnownRat (DivRat a b) => r) -> r
withDivRat = ... -- unsafeCoerce magic
```

src/Time/Units.hs
```diff
@@ -49,7 +49,7 @@ import Time.Rational (type (%), type (**), type (//), DivRat, KnownRat, Rat, Rat
 
 import qualified Control.Concurrent as Concurrent
 
-newtype Time (rat :: Rat) = Time RatioNat deriving (Show, Num)
+newtype Time (rat :: Rat) = Time RatioNat deriving (Show, Num, Eq)
```
ChShersh: I guess it's also ok to add `Ord`, `Enum`, `Integral`, `Fractional`, `Real` and `RealFrac` instance. Our `Time` newtype is just a wrapper for `Ratio` type so we can everything that `Ratio` has:

* http://hackage.haskell.org/package/base-4.10.1.0/docs/Data-Ratio.html

Maybe even have `Data` and `Storable`, but I don't know how much use we can have in such instances... Let's future issues decide :)

Also, I just discovered, that we don't need `intPart` function. I'm not aware of all numeric functions... But looks like `intPart` [is just `floor` function from `RealFrac` type class](http://hackage.haskell.org/package/base-4.10.1.0/docs/Prelude.html#v:floor). You can try to test it in `ghci` and if so, we can remove `intPart` function :) After deriving `RealFrac`, of course.

test/Test/Time/Units.hs
```diff
@@ -0,0 +1,39 @@
...
+    defaultMain tests
+
+specTests :: IO TestTree
+specTests = testSpec "Units" spec_convertUnit
+
+spec_convertUnit :: Spec
+spec_convertUnit =
+    describe "Unit Conversion Test" $ do
+        it "11 seconds is 11000 milliseconds" $
+            convertUnit @MilliSecondUnit (Time 11 :: Second) `shouldBe` Time 11000
```
ChShersh: You can write just `11000` instead of `Time 11000` and `11 :: Second` instead of `Time 11 :: Second`.
Another possible approach is to write `convertUnit @MilliSecondUnit @SecondUnit 11`. Chose which one you like more :)

PR #8, opened 2018-01-07
[#5] Pretty `Show` instance

### LINKED ISSUE #5: Pretty 'Show' instance
Currently `Show` instance is derived. In `time-units` package, when you show `3 :: Second`, you see `3s`. In `o-clock` you see `Time (3 % 1)`. I think prettier `Show` instance can be implemented. 

I have idea like this one:

```haskell
type family ShowUnit (unit :: Rat) :: Symbol
type instance ShowUnit      SecondUnit = "s"
type instance ShowUnit MicroSecondUnit = "mcs"
... etc
```

And then:

```haskell
instance Show (Time unit) where
    show (Time rat) = -- some smart combination on pattern matching on denominator + symbolVal function
```

vrom911:
Resolves #5

### DIFF
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE TypeApplications           #-}
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE TypeOperators              #-}
+{-# LANGUAGE UndecidableInstances       #-}
 
 module Time.Units
        ( -- * Time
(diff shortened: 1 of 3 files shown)


src/Time/Units.hs
```diff
@@ -75,13 +78,33 @@ type Day         = Time DayUnit
...
+type instance ShowUnit (86400   :% 1) = "d"  -- day unit
+type instance ShowUnit (604800  :% 1) = "w"  -- week unit
+type instance ShowUnit (1209600 :% 1) = "fn" -- fortnight unit
+
+instance KnownSymbol (ShowUnit unit) => Show (Time unit) where
+    show (Time rat) = let value = show (numerator rat) ++
+                           case denominator rat of
+                               1 -> ""
+                               n -> "/" ++ show n           in
+        value ++ symbolVal (Proxy @(ShowUnit unit))
```
ChShersh: Alignment here is not looking really nice to me... Don't be afraid in introducing new variables! Personally, I would write it like this:

```haskell
instance KnownSymbol (ShowUnit unit) => Show (Time unit) where
    show (Time rat) = let numeratorStr   = show (numerator rat)
                          denominatorStr = case denominator rat of
                                                1 -> ""
                                                n -> '/' : show n
```

But it's on your judgement :slightly_smiling_face:

test/Spec.hs
```diff
@@ -11,5 +11,6 @@ main = do
     print typeSpec_Normalize
     print typeSpec_DivRat
     print typeSpec_UnitCalculation
+    print typeSpec_ShowUnits
```
ChShersh: I thought a little bit more about this. I think it's better to refactor `TypeSpec` in a way you did for `tasty` tests. Just export single function `runTypeSpecTests`, which just does all this printing. Wouldn't you mind to refactor this?

PR #9, opened 2018-01-07
[#6] Add creation functions

### LINKED ISSUE #6: Convenient interface for unit creation and conversion
Currently, to work with units and conversions, you need to write something like this:

```haskell
ghci> let sec3 = 3 :: Second
ghci> let mcs3 = convertUnit sec3 :: MicroSecond
ghci> mcs3
Time (3000000 % 1)
```

It would be nice to create units with helper functions like `sec`, `mcs` (how [it's done in `serokell-util`](https://github.com/serokell/serokell-util/blob/master/src/Serokell/Util/Time.hs)).

Also, if we rename `MicroSecond` to `MicroSecondTime` and `MicroSecondUnit` to `MicroSecond` (and the other ones) it would become possible to use `-XTypeApplications`. So previous example will have the following look:

```haskell
ghci> let sec3 = sec 3
ghci> let mcs3 = convertUnit @MicroSecond sec3
ghci> mcs3
Time (3000000 % 1)
```

What do you think about this?

vrom911:
Resolves #6

### DIFF
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -35,6 +35,21 @@ module Time.Units
        , FortnightUnit
 
        , ShowUnit
+
+        -- ** Creation helpers
+       , time
+
+       , sec
+       , ms
+       , mcs
+       , ns
+
+       , minute
+       , hour
+       , day
+       , week
+       , fortnight
+
         -- ** Functions
        , convertUnit
        , threadDelay
@@ -42,7 +57,8 @@ module Time.Units
 
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Proxy (Proxy (..))
-import Data.Ratio (denominator, numerator)
+import GHC.Natural (Natural)
+import GHC.Real (Ratio ((:%)), denominator, numerator)
 import GHC.TypeLits (KnownSymbol, Symbol, symbolVal)
 
 import Time.Rational (type (%), type (**), type (//), type (:%), DivRat, KnownRat, Rat, RatioNat, divRat,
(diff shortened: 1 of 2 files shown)


src/Time/Units.hs
```diff
@@ -100,6 +114,41 @@ instance KnownSymbol (ShowUnit unit) => Show (Time unit) where
                                                 n -> '/' : show n
                       in numeratorStr ++ denominatorStr ++ symbolVal (Proxy @(ShowUnit unit))
 
+----------------------------------------------------------------------------
+-- Creation helpers
+----------------------------------------------------------------------------
+
+sec :: Natural -> Second
```
ChShersh: It would be good to add documentation with single example for every function. Something like this one:

```haskell
-- | Creates 'Second' from given 'Natural'.
-- >>> sec 42
-- 42s :: Second
```

And then just copy-paste and change interesting parts.

src/Time/Units.hs
```diff
@@ -100,6 +114,41 @@ instance KnownSymbol (ShowUnit unit) => Show (Time unit) where
...
+sec n = Time (n % 1)
```
ChShersh: You repeat `Time (n % 1)` pattern a lot. I think you can create helper function:

```haskell
time :: Natural -> Time unit
```

and then implement every function just as `sec = time` `ms = time`, etc. Every function is just a specialization of `time` function. `time` can be exported in case people want to create their own time units.

src/Time/Units.hs
```diff
@@ -100,6 +114,41 @@ instance KnownSymbol (ShowUnit unit) => Show (Time unit) where
...
+
+ms :: Natural -> MilliSecond
+ms n = Time (n % 1)
```
ChShersh: It's better to use `:%` constructor directly here, because `%` function calculates `gcd` and normalizes rational numbers. We don't need normalization, thus, functions can be faster.

src/Time/Units.hs
```diff
@@ -100,6 +114,41 @@ instance KnownSymbol (ShowUnit unit) => Show (Time unit) where
...
+
+mcs :: Natural -> MicroSecond
+mcs n = Time (n % 1)
```
ChShersh: I think it's a good idea to add `{-# INLINE mcs #-}` after every function (replaces `mcs` with proper name` and also to `time` function. Just a suggestions to compiler.

src/Time/Units.hs
```diff
@@ -118,32 +120,73 @@ instance KnownSymbol (ShowUnit unit) => Show (Time unit) where
...
-week n = Time (n % 1)
+week = time
+{-# INLINE week #-}
 
+-- | Creates 'Fortnight' from given 'Natural'.
+-- >>> fortnight 42
+-- 42fn :: Fortnight
 fortnight :: Natural -> Fortnight
-fortnight n = Time (n % 1)
+fortnight = time
```
ChShersh: You don't need to test all these functions. But you can now use them in tests to make tests nicer :slightly_smiling_face: 
This is also some sort of testing.

PR #12, opened 2018-01-07
[#10] Add Read instance

### LINKED ISSUE #10: Implement 'Read' instance manually
It would be nice to satisfy this law:

* `read . show ≡ id`

Though this requires to implement `Read` instance manually. `Show/Read` gives us basic and naive way to serialize/deserialize data. It's not really efficient. But because it's used in a lot of places, it would be good to have it. Ideally this should be used with `readMaybe` function, like this:

```haskell
ghci> readMaybe @Second "3s"
Just 3s
ghci> readMaybe @Second "5mcs"
Nothing
```

To do this, one need to implement [`readPrec` function](https://hackage.haskell.org/package/base-4.10.1.0/docs/Text-Read.html#v:readPrec) using parser-combinators interface, provided by GHC's `base` library.

vrom911:
Resolves #10

### DIFF
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -55,11 +55,17 @@ module Time.Units
        , threadDelay
        ) where
 
+import Control.Applicative ((*>))
+import Control.Monad (unless)
 import Control.Monad.IO.Class (MonadIO, liftIO)
+import Data.Char (isDigit, isLetter)
 import Data.Proxy (Proxy (..))
 import GHC.Natural (Natural)
-import GHC.Real (Ratio ((:%)), denominator, numerator)
+import GHC.Read (Read (readPrec))
+import GHC.Real (Ratio ((:%)), denominator, numerator, (%))
 import GHC.TypeLits (KnownSymbol, Symbol, symbolVal)
+import Text.ParserCombinators.ReadP (char, munch1, option, pfail)
+import Text.ParserCombinators.ReadPrec (lift)
 
 import Time.Rational (type (%), type (**), type (//), type (:%), DivRat, KnownRat, Rat, RatioNat, divRat,
                       ratVal)
@@ -67,7 +73,7 @@ import Time.Rational (type (%), type (**), type (//), type (:%), DivRat, KnownRa
 import qualified Control.Concurrent as Concurrent
 
 newtype Time (rat :: Rat) = Time RatioNat
-    deriving (Num, Eq, Ord, Enum, Fractional, Read, Real, RealFrac)
+    deriving (Num, Eq, Ord, Enum, Fractional, Real, RealFrac)
 
 -- Units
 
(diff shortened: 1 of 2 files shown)


src/Time/Units.hs
```diff
@@ -116,6 +122,16 @@ instance KnownSymbol (ShowUnit unit) => Show (Time unit) where
                                                 n -> '/' : show n
                       in numeratorStr ++ denominatorStr ++ symbolVal (Proxy @(ShowUnit unit))
 
+instance KnownSymbol (ShowUnit unit) => Read (Time unit) where
+    readPrec = lift readP
```
ChShersh: Would be good to add explicit types to `readPrec` and `readP` functions. The API of `base`'s `Read` type class is crazy and some help would be appreciated in case somebody needs to improve this instance...

src/Time/Units.hs
```diff
@@ -116,6 +122,16 @@ instance KnownSymbol (ShowUnit unit) => Show (Time unit) where
...
+      where
+        readP = do
+            n <- munch1 isDigit
+            m <- option "1" (char '/' *> munch1 isDigit)
+            timeUnitStr <- munch1 isLetter
+            unless (timeUnitStr == symbolVal (Proxy @(ShowUnit unit))) pfail
+            pure $ Time (read n % read m)
```
ChShersh: I think you do double work here. This instance can be optimized a little bit. See, you first return `1` as a string `"1"` and only then you call `read` over string `"1"`. I think we can expect that in reality users almost never will try to parse something like `13/15mcs`. Thus returning `1` as number can gain us some performance benefits. I think your instance can be improved a little bit, if you write `let naturalP = read <$> munch1 isDigit` and then you use it like this:

```haskell
m <- option 1 (char '/' *> naturalP)
```

test/Test/Time/Units.hs
```diff
@@ -32,8 +34,23 @@ spec_convertUnit =
...
         it "3 microseconds is 3/1000000 seconds" $
-            convertUnit @SecondUnit (mcs 3) `shouldBe` Time (3 % 1000000)
+            convertUnit @SecondUnit (mcs 3) `shouldBe` Time (3 :% 1000000)
         it "7 days is 1 week" $
             convertUnit @WeekUnit (day 7) `shouldBe` 1
         it "2 fornights is 28 days" $
             convertUnit @DayUnit (fortnight 2) `shouldBe` 28
+    describe "Read Time Test" $ do
+        it "parses '42s' as 42 seconds" $
+            read @Second "42s" `shouldBe` sec 42
```
ChShersh: Well, you don't need to specify `sec` and `hour` and others here because you already specified `@Second` in type of `read`. So, I guess you can either write `read @Second "42s" shouldBe 42` or `read "42s" shouldBe sec 42`. Maybe it's better to use `@`-style for consistency.

test/Test/Time/Units.hs
```diff
@@ -32,8 +34,23 @@ spec_convertUnit =
...
+        it "fails when '42mm' is expected as seconds" $
+            evaluate (read @Second "42mm") `shouldThrow` anyException
```
ChShersh: One more test to add:
* `42ms` should fail when parses as `Second`

PR #13, opened 2018-01-08
[#4] Inroduce Timestamp and functions

### LINKED ISSUE #4: Implement 'Timestamp' and 'TimeDiff' numerics
Proposed by @int-index:

1. `Timestamp` — can be negative, has no units
2. `Time` (`TimeDiff`) — cannot be negative, has units

- [x] `timeDiff :: KnownRat unit => Timestamp -> Timestamp -> (Ordering, Time unit)`
- [x] `timeAdd :: KnownRat unit => Time unit -> Time unit -> Time unit`
- [x] `timeMul :: KnownRat unit => Ratio Natural -> Time unit -> Time unit`
- [x] `timeDiv :: KnownRat unit => Time unit -> Time unit -> Ratio Natural`

### LINKED ISSUE #14: Remove the 'Num' instance for 'Time'
`(*) :: Time unit -> Time unit -> Time unit` is an incorrect type, this is not how units work. If you multiply seconds by seconds, you have seconds squared, not seconds.

vrom911:
Resolves #4 
Resolves #14 (`Num` and  `Fraction` instances are rewritten manually with `error` on `*` and `/` operators)

### DIFF
--- a/src/Time/Timestamp.hs
+++ b/src/Time/Timestamp.hs
@@ -1,0 +1,42 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE Rank2Types                 #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+
+module Time.Timestamp
+       ( Timestamp (..)
+       , timeDiff
+       , timeAdd
+       , timeMul
+       , timeDiv
+       ) where
+
+import GHC.Real (denominator, numerator, (%))
+
+import Time.Rational (KnownRat, RatioNat)
+import Time.Units (Time (..))
+
+-- | Similar to 'Time' but has no units and can be negative.
+newtype Timestamp = Timestamp Rational
+    deriving (Show, Read, Num, Eq, Ord, Enum, Fractional, Real, RealFrac)
+
+
+-- | Returns the result of comparison of two 'Timestamp's and
+-- the 'Time' of that difference of given time unit.
+timeDiff :: KnownRat unit => Timestamp -> Timestamp -> (Ordering, Time unit)
+timeDiff (Timestamp a) (Timestamp b) =
+    let order         = compare a b
+        dif           = abs (a - b)
+        t :: RatioNat = fromIntegral (numerator dif) % fromIntegral (denominator dif)
+    in (order, Time t)
+
+-- | Returns the result of addition of two 'Time' elements.
+timeAdd :: KnownRat unit => Time unit -> Time unit -> Time unit
+timeAdd = (+)
+
+-- | Returns the result of multiplication of two 'Time' elements.
+timeMul :: KnownRat unit => RatioNat -> Time unit -> Time unit
+timeMul n (Time t) = Time (n * t)
+
+-- | Returns the result of division of two 'Time' elements.
+timeDiv :: KnownRat unit => Time unit -> Time unit -> RatioNat
+timeDiv (Time t1) (Time t2) = t1 / t2
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -62,6 +62,7 @@ import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Char (isDigit, isLetter)
 import Data.Proxy (Proxy (..))
 import GHC.Natural (Natural)
+import GHC.Prim (coerce)
 import GHC.Read (Read (readPrec))
 import GHC.Real (Ratio ((:%)), denominator, numerator, (%))
 import GHC.TypeLits (KnownSymbol, Symbol, symbolVal)
@@ -73,8 +74,8 @@ import Time.Rational (type (%), type (**), type (//), type (:%), DivRat, KnownRa
 
 import qualified Control.Concurrent as Concurrent
 
-newtype Time (rat :: Rat) = Time RatioNat
-    deriving (Num, Eq, Ord, Enum, Fractional, Real, RealFrac)
+newtype Time (rat :: Rat) = Time { unTime :: RatioNat }
+    deriving (Eq, Ord, Enum, Real, RealFrac)
 
 -- Units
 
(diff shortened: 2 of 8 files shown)


src/Time/Timestamp.hs
```diff
@@ -0,0 +1,42 @@
...
+       , timeDiv
+       ) where
+
+import GHC.Real (denominator, numerator, (%))
+
+import Time.Rational (KnownRat, RatioNat)
+import Time.Units (Time (..))
+
+-- | Similar to 'Time' but has no units and can be negative.
+newtype Timestamp = Timestamp Rational
```
ChShersh: What do you think about names `TimeStamp` and `TimeDiff`?
If we rename `Time` to `TimeDiff` it's better to have `TimeStamp` for consistency, I guess. Maybe there's sense in renaming `Time` into `TimeDiff`.

src/Time/Timestamp.hs
```diff
@@ -0,0 +1,42 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE Rank2Types                 #-}
```
ChShersh: `Rank2Types` extension looks redundant to me, but I'm not sure...

src/Time/Timestamp.hs
```diff
@@ -0,0 +1,42 @@
...
+    deriving (Show, Read, Num, Eq, Ord, Enum, Fractional, Real, RealFrac)
+
+
+-- | Returns the result of comparison of two 'Timestamp's and
+-- the 'Time' of that difference of given time unit.
+timeDiff :: KnownRat unit => Timestamp -> Timestamp -> (Ordering, Time unit)
+timeDiff (Timestamp a) (Timestamp b) =
+    let order         = compare a b
+        dif           = abs (a - b)
+        t :: RatioNat = fromIntegral (numerator dif) % fromIntegral (denominator dif)
```
ChShersh: `dif` is already normalized because `(-)` operator for `Rational` performs normalization. So you can use just `(:%)` here.

src/Time/Timestamp.hs
```diff
@@ -0,0 +1,42 @@
...
+newtype Timestamp = Timestamp Rational
+    deriving (Show, Read, Num, Eq, Ord, Enum, Fractional, Real, RealFrac)
+
+
+-- | Returns the result of comparison of two 'Timestamp's and
+-- the 'Time' of that difference of given time unit.
+timeDiff :: KnownRat unit => Timestamp -> Timestamp -> (Ordering, Time unit)
+timeDiff (Timestamp a) (Timestamp b) =
+    let order         = compare a b
+        dif           = abs (a - b)
```
int-index: Let's not have `abs` here, I was thinking about something like this:

```
let
  order = compare a b
  d = case order of
    EQ -> 0
    LT -> a - b
    GT -> b - a
in
  (order, d)
```

src/Time/Timestamp.hs
```diff
@@ -0,0 +1,42 @@
...
+        t :: RatioNat = fromIntegral (numerator dif) % fromIntegral (denominator dif)
+    in (order, Time t)
+
+-- | Returns the result of addition of two 'Time' elements.
+timeAdd :: KnownRat unit => Time unit -> Time unit -> Time unit
+timeAdd = (+)
```
int-index: How did you manage to use `(+)` here? Time *most definitely* cannot have a `Num` instance.

int-index: https://github.com/serokell/o-clock/issues/14

ChShersh: Personally, I'm okay with `Num` instance for `Time`. You can write `(3 :: Second) + 5` without problems. Also you can write `(5 :: Second) - 3`. If you out of range while subtracting, when `Natural` will throw exception. Not the perfect behavior, but still might be convenient, if you use it carefully.

`(3 :: Second) * (5 :: Second)` probably doesn't make sense from physics point of view. But let's user decide what they want: more convenient interface with `(*)` or more type-safe with `timeAdd`. Also, having `Num` instances allows us to have other numeric instances with `floor` functions which are useful in real life.

int-index: `(3 :: Second) * (5 :: Second)` is ridiculous and no sane library for units would allow this.

int-index: if you insist on having a `Num` instance for numeric literals, make multiplication and division throw an error.

ChShersh: `time-units` allows this. And we never had problems with it (because we never used `*` for time units). It's just the `Num` hierarchy which is not perfect. Unfortunately, we can't have everything from `Num` except `(*)`. But the rest of `Num` looks really useful and convenient. So ¯\_(ツ)_/¯.

Having `error` inside `(*)` and `(/)` operators is okay to me.

src/Time/Units.hs
```diff
@@ -74,7 +74,7 @@ import Time.Rational (type (%), type (**), type (//), type (:%), DivRat, KnownRa
 import qualified Control.Concurrent as Concurrent
 
 newtype Time (rat :: Rat) = Time RatioNat
```
ChShersh: Maybe we can also add record field for `Time` constructor? Like

```haskell
newtype Time (rat :: Rat) = Time { unTime :: RatioNat }
```
What do you think? Sometimes can be convenient to write functions in point-free style.

src/Time/Units.hs
```diff
@@ -136,6 +136,18 @@ instance KnownSymbol (ShowUnit unit) => Read (Time unit) where
             unless (timeUnitStr == symbolVal (Proxy @(ShowUnit unit))) pfail
             pure $ Time (n % m)
 
+instance Num (Time unit) where
+    (Time t1) + (Time t2) = Time (t1 + t2)
```
ChShersh: Maybe you can add `{-# INLINE #-}` pragmas. Also you can implement all methods (except `error` ones) by delegating to `coerce` function how it's done here:

* https://hackage.haskell.org/package/base-4.10.1.0/docs/src/Data.Semigroup.html#line-243

So, this should work I guess.

```haskell
(+) = coerce ((+) :: RatioNat -> RatioNat -> RatioNat)
```

src/Time/Units.hs
```diff
@@ -136,6 +136,18 @@ instance KnownSymbol (ShowUnit unit) => Read (Time unit) where
             unless (timeUnitStr == symbolVal (Proxy @(ShowUnit unit))) pfail
             pure $ Time (n % m)
 
+instance Num (Time unit) where
```
ChShersh: I think you can also add Haddock documentation to these manual instances. Like, say, that they behave just like derived ones, except `(*)` operator throws runtime error with `error`.

PR #16, opened 2018-01-09
[#15] Add propperty-based tests for read.show=id

### LINKED ISSUE #15: Implement property-based testing with 'hedgehog' for 'read . show ≡ id'
The idea is pretty straightforward but implementation is tricky. Basically:

* Generate random rational number
* Choose random unit type for this rational number
* Convert this unit to `String` using `show`, then convert back using `read` and compare with original.

But there exist several tricky moments which I will describe below.

### How to generate rational number?

We can simply generate it using some predefined generators from `hedgehog` library, like this one:

* https://hackage.haskell.org/package/hedgehog-0.5.1/docs/Hedgehog-Gen.html#v:integral

But for our case we're interested in testing variety of cases. Specifically, rational with `denominator = 1` and not `1`. So this is the algorithm to generate rational numbers:

1. Generate random positive `numeratorValue`.
2. Generate random `Bool`.
3. If `True` then create `numeratorValue % 1` else generate random positive `denominatorValue` and create `numeratorValue % denominatorValue`.

### How to chose random unit?

We can't write function `genUnit :: Gen (Time unit)` which generates randomly `Second` or `Hour`. So the proposal is the following:

1. Introduce `GADT` type for units:

```haskell
data Unit (multiplier :: Rat) where
    USecond :: Unit SecondUnit
    UHour   :: Unit HourUnit
    -- etc.
```

2. Introduce existential data type:

```haskell
data AnyTime = forall unit . (KnownRat r, KnownSymbol (ShowUnit r)) => Time unit ::: Unit
```

Not sure about constraints. Maybe we need more or maybe some of them are redundant.

3. Generate random number from `0` to `8`.
4. Implement `unitChooser :: Int -> RatioNat -> AnyTime` function which is like

```haskell
unitChooser 0 t = Time t ::: USecond
unitChooser 1 t = Time t ::: UMilliSecond
-- etc.
```

5. Implement verifier for `AnyTime` (probably result type is `Property`? Never worked with `hedgehog`, sorry :slightly_frowning_face:)

```haskell
verifyAnyTime (t ::: USecond) = read (show t) == t
verifyAnyTime (t ::: UHour)   = read (show t) == t
-- etc.
```

vrom911:
Resolves #15

### DIFF
--- a/test/Test/Time/ReadShow.hs
+++ b/test/Test/Time/ReadShow.hs
@@ -1,0 +1,97 @@
+{-# LANGUAGE DataKinds        #-}
+{-# LANGUAGE ExplicitForAll   #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs            #-}
+{-# LANGUAGE KindSignatures   #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Time.ReadShow
+       ( readShowTestTree
+       ) where
+
+import GHC.Natural (Natural)
+import GHC.Real ((%))
+import GHC.TypeLits (KnownSymbol)
+import Hedgehog (MonadGen, MonadTest, Property, checkParallel, discover, forAll, property, (===))
+import Test.Tasty (TestTree)
+import Test.Tasty.Hedgehog (testProperty)
+
+import Time (DayUnit, FortnightUnit, HourUnit, KnownRat, MicroSecondUnit, MilliSecondUnit,
+             MinuteUnit, NanoSecondUnit, Rat, RatioNat, SecondUnit, ShowUnit, Time (..), WeekUnit)
+
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+-- | GADT for time units.
+data Unit (multiplier :: Rat) where
+    USecond      :: Unit SecondUnit
+    UMilliSecond :: Unit MilliSecondUnit
+    UMicroSecond :: Unit MicroSecondUnit
+    UNanoSecond  :: Unit NanoSecondUnit
+    UMinute      :: Unit MinuteUnit
+    UHour        :: Unit HourUnit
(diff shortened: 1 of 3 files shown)


test/Test/Time/ReadShow.hs
```diff
@@ -0,0 +1,97 @@
...
+
+-- | Generates random number from [0, 8].
+getNumber0'8 :: (MonadGen m) => m Int
+getNumber0'8 = Gen.int $ Range.constant 0 8
+
+-- | Generates random natural number up to 10^20.
+-- it receives the lower bound so that it wouldn't be possible
+-- to get 0 for denominator.
+getNatural :: (MonadGen m) => Natural -> m Natural
+getNatural n = Gen.integral (Range.constant n 100000000000000000000)
```
ChShersh: I think it's more readable to write it like `10 ^ (20 :: Int)`.

test/Test/Time/ReadShow.hs
```diff
@@ -12,55 +10,39 @@ module Test.Time.ReadShow
...
-verifyAnyTime (t ::: USecond)      = verifyProperty t
-verifyAnyTime (t ::: UMilliSecond) = verifyProperty t
-verifyAnyTime (t ::: UMicroSecond) = verifyProperty t
-verifyAnyTime (t ::: UNanoSecond)  = verifyProperty t
-verifyAnyTime (t ::: UMinute)      = verifyProperty t
-verifyAnyTime (t ::: UHour)        = verifyProperty t
-verifyAnyTime (t ::: UDay)         = verifyProperty t
-verifyAnyTime (t ::: UWeek)        = verifyProperty t
-verifyAnyTime (t ::: UFortnight)   = verifyProperty t
+verifyAnyTime (MkAnyTime t) = verifyProperty t
```
ChShersh: Probably you don't need separate `verifyProperty` function in this case since you have only single case. But that's minor.

PR #17, opened 2018-01-09
[#11] Add function examples and README

### LINKED ISSUE #11: Introduce decent documentation
To make this package more friendlier to new users it would be really good to have some documentation! What I think is good to add:

- [x] `>>>`-like examples to `convertUnit` function (3 or 4 would be good)
- [x] Three `>>>` examples how to use `threadDelay`: with `sec 2` and `2 :: Second` and `threadDelay @Second 2`
- [x] Good Readme, describing provided functionality, goals of library (how `Show` works, how `Read` works, how to create custom time unit, like `WorkWeek` (ideally with `markdown-unlit`, but not sure))

I don't think tutorial is required, library is rather small and doesn't contains a lot for now.

vrom911:
Resolves #11

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,38 @@
 [![Hackage](https://img.shields.io/hackage/v/o-clock.svg)](https://hackage.haskell.org/package/o-clock)
 [![Build status](https://travis-ci.org/serokell/o-clock.svg?branch=master)](https://travis-ci.org/serokell/o-clock)
 [![MIT license](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/serokell/o-clock/blob/master/LICENSE)
+
+`O'Clock` is the library that provides type-safe time units data types.
+
+## Overview
+
+Goals, who would use it, who would it be useful for.
+
+## Features
+
+## Example: How to make your own time unit
+
+### Setting up
+
+```haskell
+
+{-# LANGUAGE DataKinds     #-}
+{-# LANGUAGE TypeOperators #-}
+
+module Main where
+
+import Time (HourUnit, type (*))
+
+main :: IO ()
+main = putStrLn "Hello, O'Clock"
+
+-- | Time unit for a working day (8 hours).
+type WorkDayUnit = 8 * HourUnit
+-- | Time unit for a work week (5 workind days).
+type WorkWeekUnit = 5 * WorkDayUnit
+
+type WorkDay  = Time WorkDayUnit
+type WorkWeek = Time WorkWeekUnit
+
+
+```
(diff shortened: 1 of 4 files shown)


src/Time/Units.hs
```diff
@@ -244,15 +244,36 @@ fortnight = time
...
+{- | Converts from one time unit to another time unit.
+
+>>> convertUnit @HourUnit (120 :: Minute)
+2h
+
+>>> convertUnit @SecondUnit (ms 7)
+7/1000s
+
+>>> convertUnit @WeekUnit (Time @DayUnit 45)
+45/7w
```
int-index: Can we display this as `6+3/7w`?

int-index: Ah, I see, there is an issue about it already https://github.com/serokell/o-clock/issues/19

README.md
```diff
@@ -4,37 +4,127 @@
...
+```
 
 ## Features
 
+`O'Clock` provides the following features to its users:
+
+1. Single data type for all time units.
+
+   * Different time units represented as different type parameters for single `Time` data type.
+     Amount if required boilerplate is minimal.
```
vrom911: typo, "*of* required"

README.md
```diff
@@ -4,37 +4,127 @@
...
+2. Time stored as `Rational` number.
+
+   * It means that if you convert `900` milliseconds to seconds, you will have `0.9` second instead of `0` seconds.
+     So property `convertUnit @to @from . converUnit @from @to ≡ id` is satisfied.
+
+3. Different unit types are stored as rational multiplier in type.
+
+   * `o-clock` package introduces its own kind `Rat` for type-level rational numbers.
+     Units are stored as rational multipliers in type. Because of that some computation is performed on type-level.
+     So if you want to convert `Week` to `Day`, `o-clock` library ensures that time units will just be multipled by `5`.
```
vrom911: 7 days here, it's not work week from the example :smile:

README.md
```diff
@@ -4,37 +4,127 @@
...
+
+4. `base` versions of functions which work with time converted to more time-safe versions:
```
vrom911: Probably better "Functions from `base` that work with time are concerted.."

README.md
```diff
@@ -4,37 +4,127 @@
...
+### Introduce custom units
 
+You need to write some code in order to introduce your own time units. In our task we need
+work day represented as `8` hours and work week represented as `5` work days.
+
+```haskell
 -- | Time unit for a working day (8 hours).
 type WorkDayUnit = 8 * HourUnit
+
 -- | Time unit for a work week (5 workind days).
```
vrom911: Here was a typo "workin*g*", sorry (

PR #22, opened 2018-01-10
[#18] Add more functions

### LINKED ISSUE #18: Implement more time-safe functions from 'base'
`threadDelay` is probably most useful function in `base` which can use advantages of `o-clock` library. But there exist some more functions which we can rewrite into time-safe versions:

- [x] [`getCPUTime`](http://hackage.haskell.org/package/base-4.10.1.0/docs/System-CPUTime.html#v:getCPUTime)
- [x] [`timeout`](https://hackage.haskell.org/package/base-4.10.1.0/docs/System-Timeout.html)

vrom911:
Resolves #18

### DIFF
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -18,6 +18,7 @@ module Time.Units
        , MilliSecond
        , MicroSecond
        , NanoSecond
+       , PicoSecond
        , Minute
        , Hour
        , Day
@@ -29,6 +30,7 @@ module Time.Units
        , MilliSecondUnit
        , MicroSecondUnit
        , NanoSecondUnit
+       , PicoSecondUnit
        , MinuteUnit
        , HourUnit
        , DayUnit
@@ -44,6 +46,7 @@ module Time.Units
        , ms
        , mcs
        , ns
+       , ps
 
        , minute
        , hour
@@ -54,6 +57,7 @@ module Time.Units
         -- ** Functions
        , convertUnit
        , threadDelay
+       , getCPUTime
        ) where
 
 import Control.Applicative ((*>))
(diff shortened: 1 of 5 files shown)


src/Time/Units.hs
```diff
@@ -258,3 +274,14 @@ threadDelay :: forall unit m .
...
+-- | Computation getCPUTime returns the CPU time used by the current program
+-- in the given time unit.
+-- The precision of this result is implementation-dependent.
+--
+-- >>> getCPUTime @SecondUnit
+-- 1064046949/1000000000s
+getCPUTime :: forall unit m .
+              (KnownRat unit, KnownRat (PicoSecondUnit / unit), MonadIO m)
+           => m (Time unit)
+getCPUTime = liftIO CPUTime.getCPUTime >>= pure . convertUnit . ps . fromInteger
```
ChShersh: `action >>= pure . f` can be rewritten using `fmap`: `f <$> action`

src/Time/Units.hs
```diff
@@ -285,3 +287,23 @@ getCPUTime :: forall unit m .
...
+
+>>> timeout (mcs 1) (putStrLn "Hello O'Clock")
+HellNothing
+
+-}
+timeout :: forall unit m a . (MonadIO m, KnownRat unit, KnownRat (unit / MicroSecondUnit))
+        => Time unit   -- ^ time
+        -> IO a        -- ^ 'IO' action
+        -> m (Maybe a) -- ^ returns 'Nothing' if no result is available within the given time
+timeout t = liftIO . Timeout.timeout (floor $ convertUnit @MicroSecondUnit t )
```
ChShersh: This can be written in point-free style like:

```haskell
timeout = liftIO . Timeout.timeout . floor . convertUnit @MicroSecondUnit
```

With `-XTypeApplications` you can partially apply types as well.

vrom911: Unfortunately, it can't be done so easily because there is another argument which is eta-reduced now

ChShersh: Ahh, sorry, my bad :slightly_frowning_face:

PR #26, opened 2018-01-10
[#20] Refactor unit names

### LINKED ISSUE #20: Change units names
It should be `Millisecond` instead of `MilliSecond` (as I learned [here](https://www.reddit.com/r/haskell/comments/1z4oqw/tiempo_a_library_for_time_units_and_intervals/cfqnf64/)). Rename others as well.

vrom911:
Resolves #20
And resolves some haddock issue/warnings.

### DIFF
--- a/src/Time/Rational.hs
+++ b/src/Time/Rational.hs
@@ -7,13 +7,19 @@
 {-# LANGUAGE TypeOperators        #-}
 {-# LANGUAGE UndecidableInstances #-}
 
+-- | This module introduces 'Rat' kind and all necessary functional.
+
 module Time.Rational
        ( Rat (..)
        , type (:%)
+       , type (%)
        , type (*)
        , type (/)
+       , MulK
+       , DivK
        , Gcd
        , Normalize
+       , DivRat
 
         -- Utilities
        , RatioNat
@@ -34,7 +40,7 @@ import qualified GHC.TypeNats
 -- to zero.
 data Rat = Nat ::% Nat
 
--- | The result kind of overloaded multiplication and division.
+-- | The result kind of overloaded multiplication.
 type family MulK (k1 :: Type) (k2 :: Type) :: Type
 
 type instance MulK Nat Nat = Nat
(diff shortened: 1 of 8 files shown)


src/Time/Rational.hs
```diff
@@ -7,13 +7,19 @@
 {-# LANGUAGE TypeOperators        #-}
 {-# LANGUAGE UndecidableInstances #-}
 
+-- | This module introduces 'Rat' data and all necessary functional.
```
ChShersh: Better to say `Rat kind`.

src/Time/Rational.hs
```diff
@@ -160,10 +167,11 @@ type RatioNat = Ratio Natural
 
 -- | This class gives the integer associated with a type-level rational.
 class KnownRat (r :: Rat) where
-  ratVal :: Proxy r -> Ratio Natural
+    ratVal :: Proxy r -> Ratio Natural
 
 instance (KnownNat a, KnownNat b) => KnownRat (a :% b) where
     ratVal _ = natVal (Proxy @a) :% natVal (Proxy @b)
 
+-- | Returns 'Proxy' of the result of division of 'Rat's.
```
ChShersh: Triple `of`... Maybe better:

```
Returns 'Proxy' with the result of 'Rat'\'s division.
```

src/Time/TimeStamp.hs
```diff
@@ -1,6 +1,9 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
 
+-- | This module introduces 'TypeStamp' data type
```
ChShersh: Maybe `TimeStamp` and not `TypeStamp`? :slightly_smiling_face:

src/Time/Units.hs
```diff
@@ -81,6 +84,7 @@ import qualified Control.Concurrent as Concurrent
 import qualified System.CPUTime as CPUTime
 import qualified System.Timeout as Timeout
 
+-- | Time stored as rational number.
```
ChShersh: I think it's better to enhance documentation and say that time unit is represented as type level rational multiplier with kind `Rat`.

PR #28, opened 2018-01-10
[#24] Rename convertUnit to toUnit

### LINKED ISSUE #24: Invent shorter name for 'convertUnit'
In README I wrote:

```haskell
completeDays  = workDay  $ floor $ convertUnit @WorkDayUnit  workHours - convertUnit completeWeeks
```

`convertUnit` is too long name... We need more convenient and shorter way to convert units.

vrom911:
Resolves #24

### DIFF
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -58,7 +58,7 @@ module Time.Units
        , fortnight
 
         -- ** Functions
-       , convertUnit
+       , toUnit
        , threadDelay
        , getCPUTime
        , timeout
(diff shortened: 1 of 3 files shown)


src/Time/Units.hs
```diff
@@ -271,19 +271,19 @@ fortnight = time
 ----------------------------------------------------------------------------
 
 -- | Converts from one time unit to another time unit.
-convertUnit :: forall (unitTo :: Rat) (unitFrom :: Rat) .
+toUnit :: forall (unitTo :: Rat) (unitFrom :: Rat) .
                (KnownRat unitTo, KnownRat unitFrom, KnownRat (unitFrom / unitTo))
             => Time unitFrom
```
ChShersh: Probably you need to align type here as well according to our style guide.

PR #31, opened 2018-01-12
[#25] Add floorUnit function

### LINKED ISSUE #25: More convenient rounding interface
In README I wrote:

```haskell
workWeek :: Natural -> WorkWeek
workWeek = time

completeWeeks = workWeek $ floor $ convertUnit @WorkWeekUnit workHours
```

Here you can see some redundancy, because I specified `WorkWeek` twice. This is not satisfying. Problem is in type of `floor` function:

```haskell
floor :: (Integral b, RealFrac a) => a -> b
```

I think we need to do something with it...

vrom911:
Resolves #25

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Since this tutorial is literate haskell file, let's first write some pragmas and
 module Main where
 
 import Numeric.Natural (Natural)
-import Time ((:%), Time, Hour, HourUnit, UnitName, type (*), toUnit, time)
+import Time ((:%), Time, Hour, HourUnit, UnitName, type (*), floorUnit, toUnit, time)
 ```
 
 ### Introduce custom units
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -114,8 +114,8 @@ convert them to work weeks and work days and then show in human readable format.
 calculateWork :: Hour  -- type synonym for 'Time HourUnit'
               -> (WorkWeek, WorkDay)
 calculateWork workHours =
-    let completeWeeks = workWeek $ floor $ toUnit @WorkWeekUnit workHours
-        completeDays  = workDay  $ floor $ toUnit @WorkDayUnit  workHours - toUnit completeWeeks
+    let completeWeeks = floorUnit $ toUnit @WorkWeekUnit workHours
```
ChShersh: I think you can also delete `workDay` and `workWeel` smart constructors from README since they are not used anymore. It will give impression that adding new unit is even simpler :slightly_smiling_face:

src/Time/Units.hs
```diff
@@ -266,6 +267,10 @@ fortnight :: Natural -> Fortnight
 fortnight = time
 {-# INLINE fortnight #-}
 
+-- | Similar to 'floor', but works with 'Time' units.
```
ChShersh: I would like to have several examples in Haddock how this function works. These examples also can be put into unit tests.

PR #35, opened 2018-01-13
Introduce `withDivRat`

### LINKED ISSUE #7: Decide what to do with 'KnownRat (DivRat m n)' constraint
See details in this comment by @int-index:

* https://github.com/serokell/o-clock/pull/2#discussion_r160022138

vrom911:
Resolves #7

### DIFF
--- a/src/Time/Rational.hs
+++ b/src/Time/Rational.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE DataKinds            #-}
+{-# LANGUAGE AllowAmbiguousTypes  #-}
 {-# LANGUAGE TypeInType           #-}
 {-# LANGUAGE FlexibleInstances    #-}
 {-# LANGUAGE ScopedTypeVariables  #-}
@@ -24,7 +24,6 @@ module Time.Rational
         -- Utilities
        , RatioNat
        , KnownRat (..)
-       , divRat
        ) where
 
 import Data.Kind (Type)
@@ -167,11 +166,7 @@ type RatioNat = Ratio Natural
 
 -- | This class gives the integer associated with a type-level rational.
 class KnownRat (r :: Rat) where
-    ratVal :: Proxy r -> Ratio Natural
+    ratVal :: RatioNat
 
 instance (KnownNat a, KnownNat b) => KnownRat (a :% b) where
-    ratVal _ = natVal (Proxy @a) :% natVal (Proxy @b)
-
--- | Returns 'Proxy' with the result of 'Rat' \'s division.
-divRat :: forall (m1 :: Rat) (m2 :: Rat) . Proxy m1 -> Proxy m2 -> Proxy (DivRat m1 m2)
-divRat _ _ = Proxy
+    ratVal = natVal (Proxy @a) :% natVal (Proxy @b)
--- a/src/Time/Rational.hs
+++ b/src/Time/Rational.hs
@@ -1,6 +1,8 @@
 {-# LANGUAGE AllowAmbiguousTypes  #-}
 {-# LANGUAGE TypeInType           #-}
+{-# LANGUAGE FlexibleContexts     #-}
 {-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE Rank2Types           #-}
 {-# LANGUAGE ScopedTypeVariables  #-}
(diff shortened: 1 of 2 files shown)


src/Time/Rational.hs
```diff
@@ -170,3 +176,14 @@ class KnownRat (r :: Rat) where
...
+
+newtype KnownRatDict (unit :: Rat) r = MkKnownRatDict (KnownRat unit => r)
+
+giftRat :: forall (unit :: Rat) r . (KnownRat unit => r) -> RatioNat -> r
+giftRat given = unsafeCoerce (MkKnownRatDict given :: KnownRatDict unit r)
+{-# INLINE giftRat #-}
+
+-- | Performs action with introduced 'DivRat' constraint for rational numbers.
+withDivRat :: forall (a :: Rat) (b :: Rat) r . (KnownRat a, KnownRat b) => (KnownRat (a / b) => r) -> r
+withDivRat r = giftRat @(a / b) r (ratVal @a / ratVal @b)
```
ChShersh: Maybe also add `{-# INLINE #-}` pragmas to `withDivRat` and `toUnit` functions? I think it's a good idea in general to have `{-# INLINE #-}` pragmas for all our functions, like `threadDelay`, `getCPUTime`, `timeout`, etc.

src/Time/Units.hs
```diff
@@ -302,10 +301,10 @@ floorUnit = time . floor
 
 -}
 toUnit :: forall (unitTo :: Rat) (unitFrom :: Rat) .
-          (KnownRat unitTo, KnownRat unitFrom, KnownRat (unitFrom / unitTo))
+          (KnownRat unitTo, KnownRat unitFrom)
        => Time unitFrom
        -> Time unitTo
-toUnit Time{..} = Time $ unTime * ratVal (divRat (Proxy @unitFrom) (Proxy @unitTo))
+toUnit Time{..} = Time $ unTime * withDivRat @unitFrom @unitTo (ratVal @(unitFrom / unitTo))
```
int-index: I did not realize that this was the only use case for `withDivRat`. Why not:

```
toUnit (Time t) = Time (t * ratVal @unitFrom / ratVal @unitTo)
```

and then the `withDivRat` machinery won't be necessary?

ChShersh: @int-index The idea I had is that `unitFrom / unitTo` division is performed during compile-time, so when you convert `Week` to `Day` you only multiple `unTime * (5 :%1)` in runtime instead of `unTime * (BIG_NUMBER :% 1) / (ANOTHER_BIG_NUMBER :% 1)`.

int-index: Then the `reflection`-style solution we have obviously won't work, it performs division at runtime too. We'll have to keep the redundant constraints so that GHC reduces them at compile time.

ChShersh: @int-index Ok, I see. So we really need GHC plugin to infer those constraints in order to do this divistion magic at compile-time. 

The current problem we have is that we can't write `toUnit @from @to . toUnit @to @from ≡ id` testing property easily when we have such constraint in type.

Maybe you can suggest a way to implement this property and have `DivRat` constraint in `toUnit` function?

If we implement `toUnit` in simpler way, it means, that we don't need `Gcd`, type level division and all that stuff we did. This makes me feel useless and all our work kinda useless as well :disappointed:

int-index: >  So we really need GHC plugin to infer those constraints in order to do this divistion magic at compile-time.

No, the plugin would do the same thing as `reflection`, just without an explicit combinator like `withDivRat`.

> Maybe you can suggest a way to implement this property and have DivRat constraint in toUnit function?

Use `withDivRat` in the test, but not in `toUnit`?

int-index: Oh, and rename `withDivRat` to something like `withRuntimeDivRat`, so it's clear that there's a potential performance penalty.

ChShersh: > Use withDivRat in the test, but not in toUnit?

Yes, I realized this just after I wrote this message :slightly_smiling_face: So `withDivRat` combinator is useful. We can use it in tests. But type of `toUnit` and other functions should remain as it was before. Thus, we keep `DivRat` constraints in our library functions.

int-index: Yeah, this plan sounds good to me.

PR #37, opened 2018-01-13
[#34] Fix Travis CI

### LINKED ISSUE #34: Configure CI properly
Looks like CI is broken for some time. I think we should reconfigure it properly.

ChShersh:
Resolves #34


cabal.project.local
```diff
@@ -0,0 +1,9 @@
+with-compiler: /opt/ghc/8.4.1/bin/ghc
```
int-index: I don't think that's a good idea to push `cabal.project.local`, it's called `local` for a reason, it must be in `.gitignore`. Instead you could name this file `cabal.project.local.ubuntu-example`, and then `cp` in the Travis script.

ChShersh: @int-index Okay, I can do `cp`. I tried to generate it with `echo` but failed :disappointed: 
If it passes, renaming I can do.

ChShersh:
:sweat_drops:

PR #38, opened 2018-01-13
[#30] Add property tests for toUnit . toUnit = id

### LINKED ISSUE #30: Test 'toUnit @to @from . toUnit @from @to ≡ id' property with Hedgehog
We should use similar approach like we used in issue #15.

vrom911:
Resolves #30


test/Test/Time/Property.hs
```diff
@@ -41,10 +51,21 @@ unitChooser 8 t = MkAnyTime $ Time @WeekUnit        t
...
 unitChooser _ _ = error "Impossible happened"
 
--- | Verifier for 'AnyTime'.
+-- | Verifier for 'AnyTime' @read . show = id@.
 verifyAnyTime :: (MonadTest m) => AnyTime -> m ()
 verifyAnyTime (MkAnyTime t) = read (show t) === t
 
+-- | Verifier for 'toUnit'.
+verifyToUnit :: forall m . (MonadTest m) => AnyTime -> AnyTime -> m ()
+verifyToUnit (MkAnyTime t1) (MkAnyTime t2) = asd t1 t2
```
ChShersh: Maybe better to rename from `asd`? :slightly_smiling_face:

test/Test/Time/Property.hs
```diff
@@ -41,10 +51,21 @@ unitChooser 8 t = MkAnyTime $ Time @WeekUnit        t
...
+  where
+    asd :: forall (a :: Rat) (b :: Rat) .
+           (KnownSymbol (UnitName a), KnownRat a, KnownRat b)
+        => Time a
+        -> Time b
+        -> m ()
+    asd t _ = withRuntimeDivRat @b @a (toUnit @a (withRuntimeDivRat @a @b $ toUnit @b t)) === t
```
ChShersh: I think it's better to rearrange functions applications here and move both `withRuntimeDivRat` into beginning. Is it possible? It makes property more readable.

test/Test/Time/Property.hs
```diff
@@ -41,10 +51,21 @@ unitChooser 8 t = MkAnyTime $ Time @WeekUnit        t
...
--- | Verifier for 'AnyTime'.
+-- | Verifier for 'AnyTime' @read . show = id@.
 verifyAnyTime :: (MonadTest m) => AnyTime -> m ()
 verifyAnyTime (MkAnyTime t) = read (show t) === t
 
+-- | Verifier for 'toUnit'.
+verifyToUnit :: forall m . (MonadTest m) => AnyTime -> AnyTime -> m ()
+verifyToUnit (MkAnyTime t1) (MkAnyTime t2) = asd t1 t2
+  where
+    asd :: forall (a :: Rat) (b :: Rat) .
```
ChShersh: Maybe better to rename type variables to `unitTo` and `unitFrom` to make this function clearer.

test/Test/Time/Property.hs
ChShersh: Also, I think it's enough to specify only type of `unitTo` once. Type inference will do the rest of the work. Just

```haskell
toUnit (toUnit @unitTo t) === t
```

test/Test/Time/Property.hs
```diff
@@ -70,3 +91,11 @@ prop_readShowUnit = property $ do
     ratN <- forAll getRationalNum
     n    <- forAll getNumber0'9
     verifyAnyTime $ unitChooser n ratN
+
+prop_toUnit :: Property
+prop_toUnit = property $ do
+    ratN1 <- forAll getRationalNum
+    n1    <- forAll getNumber0'9
```
ChShersh: It's better to move

```haskell
     ratN <- forAll getRationalNum
     n    <- forAll getNumber0'9
    -- and unitChooser here as well
```

into separate generator, like `anyTime` since you use it already three times. Better to remove code duplication.

Also, by coding convention, it's better to not add `get` or `gen` prefixes. I see that in `hedgehog` library itself all generators don't have `getPrefix`. So maybe it's better if we follow this library naming convention in out tests as well.

test/Test/Time/Property.hs
```diff
@@ -57,45 +58,47 @@ verifyAnyTime (MkAnyTime t) = read (show t) === t
...
+rationalNum = do
+    numeratorVal <- natural 0
     isOne        <- Gen.bool
     denomVal     <- if isOne then pure 1
-                             else getNatural 1
+                             else natural 1
     return $ numeratorVal % denomVal
 
+numRatPair :: (Monad m) => PropertyT m (Int, RatioNat)
+numRatPair = liftA2 (,) (forAll number0'9) (forAll rationalNum)
```
ChShersh: You always call `unitChooser` function on the both arguments of pair. So it's even better to call `unitChooser` directly here and make this function have name `anyTime` and return `PropertyT m AnyTime`.

test/Test/Time/Property.hs
```diff
@@ -57,45 +58,47 @@ verifyAnyTime (MkAnyTime t) = read (show t) === t
...
+
 -- | Property test.
 prop_readShowUnit :: Property
 prop_readShowUnit = property $ do
-    ratN <- forAll getRationalNum
-    n    <- forAll getNumber0'9
+    (n, ratN) <- numRatPair
```
ChShersh: After you apply changes to `anyTime` function you can write this property simply as:

```haskell
property $ anyTime >>= verifyAnyTime
```

test/Test/Time/Property.hs
```diff
@@ -88,17 +87,15 @@ rationalNum = do
                              else natural 1
     return $ numeratorVal % denomVal
 
-numRatPair :: (Monad m) => PropertyT m (Int, RatioNat)
-numRatPair = liftA2 (,) (forAll number0'9) (forAll rationalNum)
```
ChShersh: You could write this as `liftA2 unitChooser (forAll number0'9) (forAll rationalNum)` if you like it more. But current implementation is good to me. And it's quite readable which is good.

test/Test/Time/Property.hs
```diff
@@ -88,17 +87,15 @@ rationalNum = do
...
-    (n, ratN) <- numRatPair
-    verifyAnyTime $ unitChooser n ratN
+prop_readShowUnit = property $ anyTime >>= verifyAnyTime
 
 prop_toUnit :: Property
-prop_toUnit = property $ do
-    (n1, ratN1) <- numRatPair
-    (n2, ratN2) <- numRatPair
-    verifyToUnit (unitChooser n1 ratN1) (unitChooser n2 ratN2)
+prop_toUnit = property $ anyTime >>= \t1 -> anyTime >>= verifyToUnit t1
```
ChShersh: This is not readable to me. Mix of `>>=` with variable names and partial application doesn't really improves things... This looks better to me:

```haskell
t1 <- anyTime
t2 <- anyTime
verityToUnit t1 t2
```

To write this shorter you need combinator like this one

```haskell
Monad m => (a -> b -> m c) -> m a -> m b -> m c
```
It's very useful but I'm not aware of such combinators :(
So I propose to write this in `do`-notation.

PR #40, opened 2018-01-13
[#32] Use type equalities in type signatures

### LINKED ISSUE #32: Use type equalities in type signatures
I have an idea. If we change types of functions from what we currently have:

```haskell
threadDelay :: forall unit m .
               (KnownRat unit, KnownRat (unit / MicrosecondUnit), MonadIO m)
            => Time unit
            -> m ()
```

to something like that:

```haskell
threadDelay :: forall time unit m .
               (time ~ Time unit, KnownRat unit, KnownRat (unit / MicrosecondUnit), MonadIO m)
            => time
            -> m ()
```

We than can run our functions like

```haskell
threadDelay @Second
```

instead of

```haskell
threadDelay @SecondUnit
```

What do you think? @vrom911 @int-index

vrom911:
Resolves #32

### DIFF
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -268,7 +268,7 @@ fortnight = time
 
 {- | Similar to 'floor', but works with 'Time' units.
 
->>> floorUnit @DayUnit (Time $ 5 % 2)
+>>> floorUnit @Day (Time $ 5 % 2)
 2d
 
 >>> floorUnit (Time @SecondUnit $ 2 % 3)
@@ -278,7 +278,7 @@ fortnight = time
 42ps
 
 -}
-floorUnit :: Time unit -> Time unit
+floorUnit :: forall time unit . (time ~ Time unit) => time -> time
 floorUnit = time . floor
 
 ----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


src/Time/Units.hs
```diff
@@ -287,22 +287,23 @@ floorUnit = time . floor
 
 {- | Converts from one time unit to another time unit.
 
->>> toUnit @HourUnit (120 :: Minute)
+>>> toUnit @Hour (120 :: Minute)
 2h
 
->>> toUnit @SecondUnit (ms 7)
+>>> toUnit @Second (ms 7)
```
ChShersh: I think we can add more example to documentation. Something like:

```haskell
>>> toUnit @Second @Minute 2
```

test/Test/Time/Units.hs
```diff
@@ -57,6 +56,6 @@ spec_Units = do
         it "returns 0s when floor < 1 second" $
             floorUnit (Time @SecondUnit $ 2 :% 3) `shouldBe` 0
         it "returns 2d when floor 2.5 days" $
-             floorUnit @DayUnit (Time $ 5 :% 2) `shouldBe` 2
+            floorUnit @DayUnit (Time $ 5 :% 2) `shouldBe` 2
```
ChShersh: I think you can also change type of `floorUnit` function.

PR #44, opened 2018-01-15
[#19] Add Formatting module

### LINKED ISSUE #19: Descending time formatting
It would be good to have some formatting function which can format time units per unit type. For example, if you have `1 day and 1500 ms)` this should be formatted as `1d1s500ms`.

This can be helpful for something like YouTrack time formatting.

Since we don't know the biggest unit for given `Time unit` the only way to implement this is as I can see: try `FortnightUnit`, then try `WeekUnit` then try `DayUnit` and so on. When we try unit, we're calling `floor` function. And we should try until `t == floor t`.

The only problem is fractions. Like `2 % 7`. I propose to format something like `2/7ms` as `285mcs714ns`.

The nicest API is to give explicitly list of units as a type to formatting function. Something like this: 

```haskell
seriesF @'[WorkWeekUnit, DayUnit, HourUnit, MinuteUnit] (minute 4000)
```

should give formatting for YouTrack like this one: `2d18h40m`

But I need to think how to implement such behavior...

vrom911:
Resolves #19 .

@ChShersh  for the moment it adds 0 timing everywhere (it could be in the beginning `0h4m2s`, in the middle `1h0m5s`, at the end `1h0m0s`). Could you please specify what interface do you prefer, or probably should make separate function on each case, so users can choose suitable for them behavior?

### DIFF
--- a/src/Time/Formatting.hs
+++ b/src/Time/Formatting.hs
@@ -1,0 +1,38 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE InstanceSigs        #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE Rank2Types          #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
+{-# LANGUAGE TypeOperators       #-}
+
+-- | This module introduces function to format time in desired way.
+--
+-- Here will be some examples after design is approved.
+
+module Time.Formatting
+       ( Series (..)
+       ) where
+
+import Time.Rational (KnownRat, Rat, withRuntimeDivRat)
+import Time.Units (KnownUnitSymbol, Time, floorUnit, toUnit)
+
+-- | Class for time formatting.
+class Series (units :: [Rat]) where
+    seriesF :: forall unit . (KnownUnitSymbol unit, KnownRat unit)
+            => Time unit
+            -> String
+
+instance Series ('[] :: [Rat]) where
+    seriesF _ = ""
+
+instance (KnownUnitSymbol unit, KnownRat unit, Series units) => Series (unit ': units :: [Rat]) where
+   seriesF :: forall unit' .
+              (KnownUnitSymbol unit', KnownRat unit')
+           => Time unit'
+           -> String
+   seriesF t = let newUnit = withRuntimeDivRat @unit' @unit $ toUnit @(Time unit) t
+                   format  = floorUnit newUnit
+               in show format ++ seriesF @units @unit (newUnit - format)
--- a/src/Time/Formatting.hs
+++ b/src/Time/Formatting.hs
@@ -8,31 +8,58 @@
 {-# LANGUAGE TypeApplications    #-}
 {-# LANGUAGE TypeOperators       #-}
 
--- | This module introduces function to format time in desired way.
---
--- Here will be some examples after design is approved.
+{- | This module introduces function to format time in desired way.
+
+__Examples__
+
+>>> seriesF @'[DayUnit, HourUnit, MinuteUnit, SecondUnit] (minute 4000)
+"2d18h40m"
+
+>>> seriesF @'[DayUnit, MinuteUnit, SecondUnit] (minute 4000)
+"2d1120m"
+
+>>>  seriesF @'[HourUnit, SecondUnit, MillisecondUnit] (Time @MinuteUnit $ 3 % 2)
+"90s"
+
+-}
 
 module Time.Formatting
        ( Series (..)
+       , unitsF
        ) where
 
-import Time.Rational (KnownRat, Rat, withRuntimeDivRat)
-import Time.Units (KnownUnitSymbol, Time, floorUnit, toUnit)
+import Time.Rational (Rat, withRuntimeDivRat)
+import Time.Units (AllUnits, KnownRatName, Time, floorUnit, toUnit)
 
 -- | Class for time formatting.
 class Series (units :: [Rat]) where
-    seriesF :: forall unit . (KnownUnitSymbol unit, KnownRat unit)
+    seriesF :: forall unit . KnownRatName unit
             => Time unit
             -> String
 
 instance Series ('[] :: [Rat]) where
+    seriesF :: Time someUnit -> String
     seriesF _ = ""
 
-instance (KnownUnitSymbol unit, KnownRat unit, Series units) => Series (unit ': units :: [Rat]) where
-   seriesF :: forall unit' .
-              (KnownUnitSymbol unit', KnownRat unit')
-           => Time unit'
+instance (KnownRatName unit, Series units) => Series (unit ': units :: [Rat]) where
+   seriesF :: forall someUnit . (KnownRatName someUnit)
+           => Time someUnit
            -> String
-   seriesF t = let newUnit = withRuntimeDivRat @unit' @unit $ toUnit @(Time unit) t
+   seriesF t = let newUnit = withRuntimeDivRat @someUnit @unit $ toUnit @(Time unit) t
                    format  = floorUnit newUnit
-               in show format ++ seriesF @units @unit (newUnit - format)
+                   timeStr = case (floor @(Time unit) newUnit :: Int) of
+                                  0 -> ""
+                                  _ -> show format
+               in timeStr ++ seriesF @units @unit (newUnit - format)
+
+{- | Similar to 'seriesF', but formats using all time units of the library.
+
+>>> unitsF $ fortnight 5
+"5fn"
+
+>>> unitsF $ minute 4000
+"2d18h40m"
+
+-}
+unitsF :: forall unit . KnownRatName unit => Time unit -> String
+unitsF = seriesF @AllUnits
--- a/src/Time/Formatting.hs
+++ b/src/Time/Formatting.hs
@@ -1,24 +1,29 @@
-{-# LANGUAGE AllowAmbiguousTypes #-}
-{-# LANGUAGE DataKinds           #-}
-{-# LANGUAGE FlexibleContexts    #-}
-{-# LANGUAGE InstanceSigs        #-}
-{-# LANGUAGE KindSignatures      #-}
-{-# LANGUAGE Rank2Types          #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications    #-}
-{-# LANGUAGE TypeOperators       #-}
+{-# LANGUAGE AllowAmbiguousTypes  #-}
+{-# LANGUAGE DataKinds            #-}
+{-# LANGUAGE FlexibleContexts     #-}
+{-# LANGUAGE InstanceSigs         #-}
+{-# LANGUAGE PolyKinds            #-}
+{-# LANGUAGE Rank2Types           #-}
+{-# LANGUAGE ScopedTypeVariables  #-}
+{-# LANGUAGE TypeApplications     #-}
+{-# LANGUAGE TypeFamilies         #-}
+{-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE UndecidableInstances #-}
 
 {- | This module introduces function to format time in desired way.
 
 __Examples__
 
->>> seriesF @'[DayUnit, HourUnit, MinuteUnit, SecondUnit] (minute 4000)
+>>> seriesF @'[Day, Hour, Minute, Second] (minute 4000)
 "2d18h40m"
 
->>> seriesF @'[DayUnit, MinuteUnit, SecondUnit] (minute 4000)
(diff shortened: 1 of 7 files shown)


ChShersh:
> or probably should make separate function on each case, so users can choose suitable for them behavior?

I think we don't need separate function for each case right now. I think we should stick to some default behavior. And if future users want something, we can add required functions :)

I'm for not printing zeros in any case. I have in mind example with YouTrack range formatting. You probably will have function like:

```haskell
ytF = seriesF @'[WorkWeek, WorkDay, Hour, Minute]
```

And you don't want to print `1d0h20m`. I think those zeros are redundant in some sense because they don't actually improve information you receive from formatted value. So let's keep formatted value minimal in it's length.

src/Time/Units.hs
```diff
@@ -135,14 +137,17 @@ type instance UnitName (86400   :% 1) = "d"  -- day unit
 type instance UnitName (604800  :% 1) = "w"  -- week unit
 type instance UnitName (1209600 :% 1) = "fn" -- fortnight unit
 
-instance KnownSymbol (UnitName unit) => Show (Time unit) where
+-- | Constraint alias for 'KnownSymbol' 'UnitName'.
+type KnownUnitSymbol unit = KnownSymbol (UnitName unit)
```
ChShersh: Both constraints `KnownUnitSymbol` and `KnownRat` also appear quite often. You can create another constraint like `KnownRatName`. Also, let's rename `KnownUnitSymbol` to `KnownUnitName` to make it shorter :slightly_smiling_face:

src/Time/Formatting.hs
```diff
@@ -0,0 +1,38 @@
...
+module Time.Formatting
+       ( Series (..)
+       ) where
+
+import Time.Rational (KnownRat, Rat, withRuntimeDivRat)
+import Time.Units (KnownUnitSymbol, Time, floorUnit, toUnit)
+
+-- | Class for time formatting.
+class Series (units :: [Rat]) where
+    seriesF :: forall unit . (KnownUnitSymbol unit, KnownRat unit)
```
ChShersh: I don't actually like duplication of `unit` type variable name I introduced here. And `unit'` is not really good name. Let's replace `unit` and `unit'` in `Time unit` with `someUnit`.

src/Time/Formatting.hs
```diff
@@ -0,0 +1,38 @@
...
+            => Time unit
+            -> String
+
+instance Series ('[] :: [Rat]) where
+    seriesF _ = ""
```
ChShersh: Let's put `seriesF` type signature here as well. I like `InstanceSigs` extension. Types can't be bad. Especially when such type-level sorcery is happening...

src/Time/Formatting.hs
```diff
@@ -0,0 +1,38 @@
...
+
+instance (KnownUnitSymbol unit, KnownRat unit, Series units) => Series (unit ': units :: [Rat]) where
+   seriesF :: forall unit' .
+              (KnownUnitSymbol unit', KnownRat unit')
+           => Time unit'
+           -> String
+   seriesF t = let newUnit = withRuntimeDivRat @unit' @unit $ toUnit @(Time unit) t
```
ChShersh: I also think it's a good idea to create type-level list consisting of all time units in our library and implement `unitsF` formatter which prints units using this list. After we remove redundant zeros, this formatter will be really helpful. I think you can put this list into `Time.Units` module to not reexport every time unit in `Formatting` module.

src/Time/Formatting.hs
```diff
@@ -8,31 +8,58 @@
...
 
 instance Series ('[] :: [Rat]) where
+    seriesF :: Time someUnit -> String
     seriesF _ = ""
 
-instance (KnownUnitSymbol unit, KnownRat unit, Series units) => Series (unit ': units :: [Rat]) where
-   seriesF :: forall unit' .
-              (KnownUnitSymbol unit', KnownRat unit')
-           => Time unit'
+instance (KnownRatName unit, Series units) => Series (unit ': units :: [Rat]) where
```
ChShersh: Okay, next step is to parametrize implementation by `Time` instead of `Rat`, so we can write
```haskell
seriesF @'[Hour]
```
instead of `seriesF @'[HourUnit]`.

test/Test/Time/Units.hs
```diff
@@ -59,3 +60,21 @@ spec_Units = do
...
+            seriesF @'[HourUnit, SecondUnit, MillisecondUnit] (Time @MinuteUnit $ 3 :% 2) `shouldBe` "90s"
+        it "works without minutes formatting" $
+            seriesF @'[DayUnit, MinuteUnit, SecondUnit] (minute 4000) `shouldBe` "2d1120m"
+
+        it "4000 minutes should be formatted like 2d18h40m" $
+            unitsF (minute 4000) `shouldBe` "2d18h40m"
+        it "42 fortnights should be formatted like 42fn" $
+            unitsF (fortnight 42) `shouldBe` "42fn"
+        it "empty when receive zero time" $
+            unitsF (Time @HourUnit 0) `shouldBe` ""
```
ChShersh: Personally I don't like this behavior. I didn't think of it. But I think that if `0` time is passed to `seriesF` formatter we want to use `0smth`. So the questions is what you would like to see in case you call:

```haskell
seriesF @'[Hour, Second] (minute 0)
```

1. `0h`
2. `0m`
3. `0s`

I think this deserves separate issue. I will create it.

test/Test/Time/Units.hs
```diff
@@ -59,3 +60,21 @@ spec_Units = do
...
+        it "4000 minutes should be formatted without beginning-zeros" $
+            seriesF @'[WeekUnit, DayUnit, HourUnit, MinuteUnit] (minute 4000) `shouldBe` "2d18h40m"
+        it "3601 sec should be formatted without middle-zeros" $
+            seriesF @'[HourUnit, MinuteUnit, SecondUnit] (sec 3601) `shouldBe` "1h1s"
+        it "works on rational nums" $
+            seriesF @'[HourUnit, SecondUnit, MillisecondUnit] (Time @MinuteUnit $ 3 :% 2) `shouldBe` "90s"
+        it "works without minutes formatting" $
+            seriesF @'[DayUnit, MinuteUnit, SecondUnit] (minute 4000) `shouldBe` "2d1120m"
+
+        it "4000 minutes should be formatted like 2d18h40m" $
```
ChShersh: Let's also create test with formatting of all units. Like `fn 1 + week 1 + day 1 + ...`.
Of course, it's not possible to add  weeks and days in our library. But I think we can create some operator like this:

```haskell
(+:) :: Time a -> Time b -> Time b
```

I personally find it useful.

vrom911:
@ChShersh I wanted to make something like `times :: [t :: (Time (unit :: Rat))]` but I couldn't implement that. What am I missing in this direction of implementation?

ChShersh:
@vrom911 I'm not sure how to achieve this behavior. We know that elements of lists will be only of type `Time unit` but I'm not sure how to convince compiler in that fact. Probably @int-index knows.

src/Time/Formatting.hs
```diff
@@ -1,24 +1,29 @@
...
-{-# LANGUAGE KindSignatures      #-}
-{-# LANGUAGE Rank2Types          #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications    #-}
-{-# LANGUAGE TypeOperators       #-}
+{-# LANGUAGE AllowAmbiguousTypes  #-}
+{-# LANGUAGE DataKinds            #-}
+{-# LANGUAGE FlexibleContexts     #-}
+{-# LANGUAGE InstanceSigs         #-}
+{-# LANGUAGE PolyKinds            #-}
```
ChShersh: I think `PolyKinds` extension is redundant here.

src/Time/Formatting.hs
```diff
@@ -28,29 +33,30 @@ module Time.Formatting
...
-                   timeStr = case (floor @(Time unit) newUnit :: Int) of
-                                  0 -> ""
-                                  _ -> show format
-               in timeStr ++ seriesF @units @unit (newUnit - format)
+instance (time ~ Time unit, KnownRatName unit, Series times)
+    => Series (time ': times :: [*]) where
+    seriesF :: forall someTime someUnit . (someTime ~ Time someUnit, KnownRatName someUnit)
+            => someTime
+            -> String
+    seriesF t = let newUnit = withRuntimeDivRat @someUnit @unit $ toUnit @(Time unit) t
```
ChShersh: I think you can use `@time` here instead of `Time unit`.

src/Time/Formatting.hs
```diff
@@ -28,29 +33,30 @@ module Time.Formatting
...
+                    format  = floorUnit newUnit
+                    timeStr = case (floor @time newUnit :: Int) of
```
ChShersh: I think `()` are redundant here. You can try to run `hlint` over our project. But I'm not sure it can parse it...

src/Time/Units.hs
```diff
@@ -131,6 +125,13 @@ type Day         = Time DayUnit
 type Week        = Time WeekUnit
 type Fortnight   = Time FortnightUnit
 
+-- | Type-level list that consist of all time.
+type AllTimes =
```
ChShersh: I think it's a good idea to add explicit kind to `AllTimes` variable here.

test/Test/Time/Units.hs
```diff
@@ -62,15 +62,15 @@ spec_Units = do
...
+            seriesF @'[Day, Hour, Minute, Second] (minute 4000) `shouldBe` "2d18h40m"
         it "4000 minutes should be formatted without beginning-zeros" $
-            seriesF @'[WeekUnit, DayUnit, HourUnit, MinuteUnit] (minute 4000) `shouldBe` "2d18h40m"
+            seriesF @'[Week, Day, Hour, Minute] (minute 4000) `shouldBe` "2d18h40m"
         it "3601 sec should be formatted without middle-zeros" $
-            seriesF @'[HourUnit, MinuteUnit, SecondUnit] (sec 3601) `shouldBe` "1h1s"
+            seriesF @'[Hour, Minute, Second] (sec 3601) `shouldBe` "1h1s"
         it "works on rational nums" $
-            seriesF @'[HourUnit, SecondUnit, MillisecondUnit] (Time @MinuteUnit $ 3 :% 2) `shouldBe` "90s"
+            seriesF @'[Hour, Second, Millisecond] (Time @MinuteUnit $ 3 :% 2) `shouldBe` "90s"
```
ChShersh: Btw, I just understood that you can create ration time using `min 3 / 2` way.

src/Time/Units.hs
```diff
@@ -299,6 +301,18 @@ fortnight = time
...
+-- | Sums times of different units.
+--
+-- >>> minute 1 +: sec 1
+-- 61s
+--
+(+:) :: forall timeB timeA b a . (timeA ~ Time a, timeB ~ Time b, KnownDivRat a b)
+     => timeA
+     -> timeB
+     -> timeB
+t1 +: t2 = (toUnit @timeB t1) + t2
```
ChShersh: `()` are redundant here. `@timeB` as well because type inference algorithm can handle this. But actually it's a good idea to keep explicit type applications here in case type of this operator will be changed in future.

int-index:
I don't really like that we started conflating `Time unit` and `unit` for the sake of slightly better `-XTypeApplications` notation. What we have above is an example of why it's clearly not working out well.

ChShersh:
@int-index 
> I don't really like that we started conflating Time unit and unit for the sake of slightly better -XTypeApplications notation. What we have above is an example of why it's clearly not working out well.

So you say that it's not possible to constrain kind of `Series` type parameter somehow to allow only `Time` data type? You think it's better for this exact type class to keep `Rat` kind? And reason to do so? Because using `Time` we have slightly better interface. Though, this interface can be achieved in another way by renaming `Second -> SecondTime` and `SecondUnit -> Second` though I don't really like this renaming because it clashes a bit with semantic of data types...

int-index:
> So you say that it's not possible to constrain kind of Series type parameter somehow to allow only Time data type? 

It's possible in two ways:

* you use a different kind for elements (`Rat` seems a perfect fit here :wink:)
* you add a constraint, `type family AllTime (ts :: [Type]) :: Constraint`, that performs the check. Of course, now you'll have these constraints floating around for no reason.

The renaming solution sounds better to me, I don't mind having `Second :: Rat` and writing `Time Second` in my code, I'd read it as "time measured in seconds", which makes sense.

int-index:
There was another solution that I experimented with, we could have

```
type family Second :: k
```

where `k` can be instantiated to either `Rat` or `Type`. The reason I decided not to do that is that it breaks overloaded multiplication/division operators.

Ideally, of course, we'd want to have `15 :: k` overloaded (just like we have `15 :: Num n => n` at term level), and then all problems would go away, but this will require a change to GHC (we could start a ghc-proposal and motivate it by our library).

src/Time/Units.hs
vrom911: It's not possible to write so `AllTimes :: [*]`..

ChShersh: Oh, sorry. I never did this and somehow thought it's possible.

PR #52, opened 2018-01-17
[#33] Add benchmarks with gauge

### LINKED ISSUE #33: Introduce simple benchmarks
We want to compare with `time-units` package in this areas:

1. Convert `1s` to nanoseconds.
2. Convert `1000ns` to seconds.
3. Convert `1w` to days.

`o-clock` most likely will be much slower. But I think it's a good starting point.

If possible, it will be good to compare with `tiempo` and `data-timeout` packages. Also, instead of `criterion` we can try to use `gauge`: 

* https://github.com/vincenthz/hs-gauge

vrom911:
Resolves #33 .

Current results I get:
screenshot from 2018-01-17 17-34-30

### DIFF
--- a/benchmark/Main.hs
+++ b/benchmark/Main.hs
@@ -1,2 +1,25 @@
+{-# LANGUAGE TypeApplications #-}
+
+module Main where
+
+import Gauge (bench, bgroup, defaultMain, whnf)
+
+import Time (Day, Nanosecond, Second, ns, sec, toUnit, week)
+
+import qualified Data.Time.Units as TU (Day, Nanosecond, Second, Week, convertUnit)
+
 main :: IO ()
-main = undefined
+main = defaultMain
+    [ bgroup "Second to Nanosecond"
+        [ bench "o'clock"    $ whnf (toUnit @Nanosecond . sec) 1
+        , bench "time-units" $ whnf (TU.convertUnit :: TU.Second -> TU.Nanosecond) 1
+        ]
+    , bgroup "1000ns to s"
+        [ bench "o'clock"    $ whnf (toUnit @Second . ns) 1000
+        , bench "time-units" $ whnf (TU.convertUnit :: TU.Nanosecond -> TU.Second) 1000
+        ]
+    , bgroup "week to days"
+        [ bench "o'clock"    $ whnf (toUnit @Day . week ) 1
+        , bench "time-units" $ whnf (TU.convertUnit :: TU.Week -> TU.Day) 1
+        ]
+    ]
--- a/benchmark/Main.hs
+++ b/benchmark/Main.hs
@@ -1,32 +1,48 @@
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
-{-# LANGUAGE DataKinds        #-}
-{-# LANGUAGE KindSignatures   #-}
-{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE DataKinds                  #-}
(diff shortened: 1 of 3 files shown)


ChShersh:
Okay, results are quite good! It's still some `Nanosecond`s. But we gain safety with `Ratio Natural`. Performance is not a bottleneck.

o-clock.cabal
```diff
@@ -84,7 +84,9 @@ benchmark o-clock-benchmark
   ghc-options:         -Wall -Werror -O2 -threaded -rtsopts -with-rtsopts=-N
```
ChShersh: Could you try to add `-O2` option to library temporary and see whether it changes something or not?

benchmark/Main.hs
```diff
@@ -1,2 +1,25 @@
...
+
+import Time (Day, Nanosecond, Second, ns, sec, toUnit, week)
+
+import qualified Data.Time.Units as TU (Day, Nanosecond, Second, Week, convertUnit)
+
 main :: IO ()
-main = undefined
+main = defaultMain
+    [ bgroup "Second to Nanosecond"
+        [ bench "o'clock"    $ whnf (toUnit @Nanosecond . sec) 1
```
ChShersh: Could you switch frm `wnhf` to `nf`? `Rational` is a complex data type that has constructor consisting of several fields.

benchmark/Main.hs
```diff
@@ -1,2 +1,25 @@
...
+        , bench "time-units" $ whnf (TU.convertUnit :: TU.Second -> TU.Nanosecond) 1
+        ]
+    , bgroup "1000ns to s"
+        [ bench "o'clock"    $ whnf (toUnit @Second . ns) 1000
+        , bench "time-units" $ whnf (TU.convertUnit :: TU.Nanosecond -> TU.Second) 1000
+        ]
```
ChShersh: Nice! It would be also good to compare with `tiempo` and `data-timeout` packages (of course, only if we can compile them).

o-clock.cabal
vrom911: It doesn't influence on the results, so I'm not adding it :slightly_smiling_face:

benchmark/Main.hs
vrom911: There is a problem with `Semigroup` instance in `data-timeout`.
Should I try to handle it within the data-timeout library and then add benchs?

ChShersh: @vrom911 I think you shouldn't. `data-timeout` library looks not maintained.

benchmark/Main.hs
```diff
@@ -1,25 +1,32 @@
...
-import Gauge (bench, bgroup, defaultMain, whnf)
+import Control.DeepSeq (NFData)
+import Gauge (bench, bgroup, defaultMain, nf, whnf)
 
-import Time (Day, Nanosecond, Second, ns, sec, toUnit, week)
+import Time (Day, Nanosecond, Rat, Second, Time, ns, sec, toUnit, week)
 
 import qualified Data.Time.Units as TU (Day, Nanosecond, Second, Week, convertUnit)
 
+instance NFData (Time (unit :: Rat))
```
ChShersh: `NFData` type class has default definition of `rnf` function which gives us behavior similar to `whnf` function. Could you try to use `StandaloneDeriving` language extension and write this instance like this:

```haskell
deriving NFData (Time (unit :: Rat))
```

PR #55, opened 2018-01-17
[#45] Fix behavior of 0 time passed to seriesF

### LINKED ISSUE #45: Decide (and fix) what to do with 0 time passed to 'seriesF'
I think that if `0` time is passed to `seriesF` formatter we want to use `0smth`. So the question is what you would like to see in case you call:

```haskell
seriesF @'[Hour, Second] (minute 0)
```

1. `0h`
2. `0m`
3. `0s`

vrom911:
Resolves #45 .

### DIFF
--- a/src/Time/Formatting.hs
+++ b/src/Time/Formatting.hs
@@ -26,6 +26,9 @@ __Examples__
 >>>  seriesF @'[Hour, Second, Millisecond] (Time @Minute $ 3 % 2)
 "90s"
 
+>>> seriesF @'[Hour, Second] (minute 0)
+"0h"
+
 -}
 
 module Time.Formatting
@@ -59,11 +62,14 @@ instance (KnownRatName unit, Series units)
 #else
     seriesF t = let newUnit = toUnit @unit t
 #endif
-                    format  = floorUnit newUnit
-                    timeStr = case floor newUnit :: Int of
+                    flooredNewUnit = floorUnit newUnit
+                    timeStr = case flooredNewUnit of
                                    0 -> ""
-                                   _ -> show format
-                in timeStr ++ seriesF @units @unit (newUnit - format)
+                                   _ -> show flooredNewUnit
+                    nextUnit = newUnit - flooredNewUnit
+                in if nextUnit == 0
+                   then show newUnit
+                   else timeStr ++ seriesF @units @unit nextUnit
 
 {- | Similar to 'seriesF', but formats using all time units of the library.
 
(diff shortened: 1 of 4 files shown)


src/Time/Formatting.hs
```diff
@@ -63,7 +66,9 @@ instance (KnownRatName unit, Series units)
                     timeStr = case floor newUnit :: Int of
                                    0 -> ""
                                    _ -> show format
-                in timeStr ++ seriesF @units @unit (newUnit - format)
+                in if t == 0 || format == newUnit
+                   then show newUnit
+                   else timeStr ++ seriesF @units @unit (newUnit - format)
```
ChShersh: I have a feeling that this could be written simpler... I think that

* `floor newUnit == 0 ≡ floorUnit newUnit == 0`

So at least one call to `floor` is redundant. Also, I think that `format` is not a good name for variable with value `floorUnit newUnit`...

src/Time/Formatting.hs
```diff
@@ -62,13 +62,13 @@ instance (KnownRatName unit, Series units)
 #else
     seriesF t = let newUnit = toUnit @unit t
 #endif
-                    format  = floorUnit newUnit
-                    timeStr = case floor newUnit :: Int of
+                    flooredNewUnit  = floorUnit newUnit
```
ChShersh: Extra space before `=` operator 🙂

src/Time/Formatting.hs
```diff
@@ -62,13 +62,13 @@ instance (KnownRatName unit, Series units)
...
+                    timeStr = case flooredNewUnit of
                                    0 -> ""
-                                   _ -> show format
-                in if t == 0 || format == newUnit
+                                   _ -> show flooredNewUnit
+                in if t == 0 || flooredNewUnit == newUnit
                    then show newUnit
-                   else timeStr ++ seriesF @units @unit (newUnit - format)
+                   else timeStr ++ seriesF @units @unit (newUnit - flooredNewUnit)
```
ChShersh: I think that instead of pattern-matching on `flooredUnit` we can pattern-match on `newUnit - flooredUnit`. I think the following holds:

* `floor newUnit == 0 ≡ newUnit - floor newUnit == 0`
* `newUnit == 0 ==> newUnit - floorUnit == 0`

So we can pattern match on the result of subtraction simplify case. Well, by simplifying I also mean remove `t == 0` check. But I don't think that `seriesF` function would be a bottleneck of any application. It's already uses `String` and polymorphic recursion. And by removing check `t == 0` we only decrease performance of showing zero time while increasing performance of showing non-zero time. I think `seriesF` would be called on non-zero times more often 🙂

PR #57, opened 2018-01-18
[#51] Verify seriesF

### LINKED ISSUE #51: Verify units order for 'seriesF' function
This call makes sense:

```haskell
seriesF @'[Hour, Minute]
```

While this one doesn't:

```haskell
seriesF @'[Minute, Hour]
```

It would be good to define some type-level predicate `IsDescending` using type families for ensuring more compiletime safety in `seriesF` function.

vrom911:
Resolves #51 .

### DIFF
--- a/src/Time/Formatting.hs
+++ b/src/Time/Formatting.hs
@@ -55,20 +55,18 @@ import Time.Rational (type (>=%), withRuntimeDivRat)
 import Time.Units (AllTimes, KnownRatName, Time, floorUnit, toUnit)
 
 #if ( __GLASGOW_HASKELL__ >= 804 )
--- Type-level 'if' for 'Bool's.
-type family If (c :: Bool) (t :: Bool) (e :: Bool) where
-  If 'True  t e = t
-  If 'False t e = e
+-- Type-level 'and'.
+type family And (a :: Bool) (b :: Bool) where
+  And 'False _ = 'False
+  And _      x = x
 
 -- | Type family for verification of the descending order of the given
 -- list of time units.
-type family IsDescending (units :: [Rat]) :: Bool
-
-type instance IsDescending (unit1 ': unit2 ': units :: [Rat]) =
-    If (unit1 >=% unit2) (IsDescending (unit2 ': units :: [Rat]))
-                         'False
-type instance IsDescending (unit ': '[] :: [Rat]) = 'True
-type instance IsDescending ('[] :: [Rat]) = 'True
+type family IsDescending (units :: [Rat]) :: Bool where
+    IsDescending ('[])     = 'True
+    IsDescending ('[unit]) = 'True
+    IsDescending (unit1 ': unit2 ': units) =
+        (unit1 >=% unit2) `And` (IsDescending (unit2 ': units))
 #endif
 
 -- | Class for time formatting.
--- a/src/Time/Formatting.hs
+++ b/src/Time/Formatting.hs
@@ -35,8 +35,7 @@ Example of the error from @ghci@:
 >>> seriesF @'[Millisecond, Second] (minute 42)
 
 <interactive>:10:2: error:
-    • Couldn't match type ‘'False’ with ‘'True’
-        arising from a use of ‘seriesF’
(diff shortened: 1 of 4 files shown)


src/Time/Formatting.hs
```diff
@@ -38,10 +50,27 @@ module Time.Formatting
...
-import Time.Rational (withRuntimeDivRat)
+import Time.Rational (type (>=%), withRuntimeDivRat)
 #endif
 import Time.Units (AllTimes, KnownRatName, Time, floorUnit, toUnit)
 
+#if ( __GLASGOW_HASKELL__ >= 804 )
+-- Type-level 'if' for 'Bool's.
+type family If (c :: Bool) (t :: Bool) (e :: Bool) where
+  If 'True  t e = t
+  If 'False t e = e
```
ChShersh: I think the code can become a little bit nicer if instead of `If` you write type level `And` function.

src/Time/Formatting.hs
```diff
@@ -38,10 +50,27 @@ module Time.Formatting
...
+
+-- | Type family for verification of the descending order of the given
+-- list of time units.
+type family IsDescending (units :: [Rat]) :: Bool
+
+type instance IsDescending (unit1 ': unit2 ': units :: [Rat]) =
+    If (unit1 >=% unit2) (IsDescending (unit2 ': units :: [Rat]))
+                         'False
+type instance IsDescending (unit ': '[] :: [Rat]) = 'True
```
ChShersh: I guess you can write it like `'[unit]`.

src/Time/Formatting.hs
```diff
@@ -38,10 +50,27 @@ module Time.Formatting
...
+type instance IsDescending ('[] :: [Rat]) = 'True
```
ChShersh: I think you don't need to specify `:: [Rat]` in every case because you already specified type of `IsDescending` family argument.

src/Time/Formatting.hs
```diff
@@ -52,7 +81,11 @@ instance Series ('[] :: [Rat]) where
     seriesF :: Time someUnit -> String
     seriesF _ = ""
 
-instance (KnownRatName unit, Series units)
+instance ( KnownRatName unit, Series units
+#if ( __GLASGOW_HASKELL__ >= 804 )
+         , IsDescending (unit : units) ~ 'True
```
ChShersh: Okay, here comes the most interesting part. As you already described in documentation, the type error user will see contains message like this:

```haskell
Couldn't match type ‘'False’ with ‘'True’
```

This error message is not actually very useful. I think it would be better to return some custom `ErrorMessage` if `IsDescending ~ 'False` like you did in `universum`:

* https://github.com/serokell/universum/blob/2c524b5ddefad5b21232aee59fcb6caf8ff1c9f6/src/Universum/Container/Class.hs#L697

src/Time/Formatting.hs
```diff
@@ -38,10 +50,27 @@ module Time.Formatting
...
 
+#if ( __GLASGOW_HASKELL__ >= 804 )
+-- Type-level 'if' for 'Bool's.
+type family If (c :: Bool) (t :: Bool) (e :: Bool) where
+  If 'True  t e = t
+  If 'False t e = e
+
+-- | Type family for verification of the descending order of the given
+-- list of time units.
+type family IsDescending (units :: [Rat]) :: Bool
```
ChShersh: This type family should be closed. See this documentation for example:

* https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#closed-type-families

src/Time/Formatting.hs
vrom911: I have no idea how is that working, so I won't be able to implement that, sorry, it's too magical to me to understand.

ChShersh: @vrom911 Don't worry about that :) I think only 1% of Haskellers (or even less) is able to implement that. And I wasn't able as well until some recent times. I agree, this is indeed to magical... I will do that after rest of your changes.

PR #59, opened 2018-01-22
[#46] Introduce ... type for lists creation

### LINKED ISSUE #46: Ability to create custom list of time units
In my thoughts, probably it would be helpful to have something to create the list with the given max and min time like `magicSmth Hour Millisecond = '[Hour, Minute, Second, Millisecond]`.

vrom911:
Resolves #46

### DIFF
--- a/o-clock.cabal
+++ b/o-clock.cabal
@@ -24,6 +24,7 @@ source-repository head
 library
   hs-source-dirs:      src
   exposed-modules:     Time
+                         Time.Bool
                          Time.Formatting
                          Time.Rational
                          Time.TimeStamp
@@ -35,6 +36,8 @@ library
   default-language:    Haskell2010
   default-extensions:  OverloadedStrings
                        RecordWildCards
+                       ScopedTypeVariables
+                       TypeFamilies
 
 executable play-o-clock
   main-is:             Playground.hs
--- a/src/Time/Bool.hs
+++ b/src/Time/Bool.hs
@@ -1,0 +1,21 @@
+{-# LANGUAGE DataKinds            #-}
+{-# LANGUAGE PolyKinds            #-}
+{-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | This module contains functions to work with type-level 'Bool's.
+
+module Time.Bool
+       ( If
+       , type (&&)
+       ) where
+
+-- | Type-level 'if'.
+type family If (x :: Bool) (a :: k) (b :: k) :: k where
+    If 'True  a _ = a
+    If 'False _ b = b
+
+-- | Type-level 'and'.
+type family (a :: Bool) && (b :: Bool) :: Bool where
+    'False && _ = 'False
+    _      && x = x
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -7,9 +7,7 @@
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE InstanceSigs               #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE TypeApplications           #-}
-{-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE TypeOperators              #-}
 {-# LANGUAGE UndecidableInstances       #-}
 
@@ -31,8 +29,6 @@ module Time.Units
        , Day
        , Week
        , Fortnight
-
-       , AllTimes
 
        , UnitName
        , KnownUnitName
@@ -125,13 +121,6 @@ type Fortnight   = 1209600 :% 1
 -- | Time unit is represented as type level rational multiplier with kind 'Rat'.
 newtype Time (rat :: Rat) = Time { unTime :: RatioNat }
     deriving (Eq, Ord, Enum, Real, RealFrac, Generic)
-
--- | Type-level list that consist of all times.
-type AllTimes =
-  '[ Fortnight, Week, Day, Hour, Minute, Second
-   , Millisecond , Microsecond, Nanosecond, Picosecond
-   ]
-
 
 -- | Type family for prettier 'show' of time units.
 type family UnitName (unit :: Rat) :: Symbol
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The change log is available [on GitHub][2].
 0.0.1
 =====
 
+* [#46](https://github.com/serokell/o-clock/issues/46):
+  Introduce `...` type to create custom time unit lists in
+  provided bounds.
 * [#51](https://github.com/serokell/o-clock/issues/51):
   Add `IsDescending` type family to check lists of time units
   in `seriesF` function on right order
--- a/src/Time/Formatting.hs
+++ b/src/Time/Formatting.hs
@@ -4,9 +4,7 @@
 {-# LANGUAGE FlexibleContexts     #-}
 {-# LANGUAGE InstanceSigs         #-}
 {-# LANGUAGE Rank2Types           #-}
-{-# LANGUAGE ScopedTypeVariables  #-}
 {-# LANGUAGE TypeApplications     #-}
-{-# LANGUAGE TypeFamilies         #-}
 {-# LANGUAGE TypeOperators        #-}
 {-# LANGUAGE UndecidableInstances #-}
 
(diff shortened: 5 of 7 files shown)


o-clock.cabal
```diff
@@ -28,6 +28,7 @@ library
                          Time.Rational
                          Time.TimeStamp
                          Time.Units
+  other-modules:       Time.Bool
```
ChShersh: Actually it's better to put all modules into `exposed-modules`. Practice shows that in reality there's no actual benefit in putting something into `other-modules` but some drawbacks are possible (bad linking, absence of documentation). So currently the practice in Haskell community is to put every module into `exposed-modules` and if you want to specify the fact that module is internal you use `Internal` word here, like, `Time.Internal.Bool`. Though I'm not sure that this module should have name `Time.Internal.Bool`. It contains `==%` useful type family people might want to use in their applications.

src/Time/Bool.hs
```diff
@@ -0,0 +1,39 @@
...
+{-# LANGUAGE ExplicitForAll             #-}
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE InstanceSigs               #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE TypeApplications           #-}
+{-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE TypeInType                 #-}
+{-# LANGUAGE TypeOperators              #-}
+{-# LANGUAGE UndecidableInstances       #-}
```
ChShersh: Okay, now I understand that 15 pragmas per module is too much even for me... Could you please move extensions we have in every module into `default-extensions` inside `.cabal` file? Also I think that `CPP` `TypeInType` `AllowAmbiguosTypes` (and probably something else) is not needed in that module... Unfortunately, GHC cannot tell which extensions are unused so this requires some manual extra work :(

src/Time/Bool.hs
```diff
@@ -0,0 +1,39 @@
...
+
+import Time.Rational (type (:%), Rat)
+
+-- | Type-level 'if'.
+type family If (x :: Bool) a b where
+    If 'True  a _ = a
+    If 'False _ b = b
+
+-- | Type-level 'and'.
+type family And (a :: Bool) (b :: Bool) where
```
ChShersh: Could you also write kinds of type family result, not only arguments? This kind can be inferred. But I think that it's better to write it explicitly, like we write types of arguments and type of result for usual functions.

src/Time/Bool.hs
```diff
@@ -0,0 +1,39 @@
...
+  And 'False _ = 'False
+  And _      x = x
```
ChShersh: What do you think about renaming `And` to `&&` operator? It might be more convenient to use (2 characters in infix usage instead of 5) but maybe this operator might look confusing... I think we need some package with type families for `Bool`. I didn't find such package on Hackage.

src/Time/Bool.hs
```diff
@@ -0,0 +1,39 @@
...
+
+-- | A type family to compute equality of 'Rat's.
+type family (==%) (m :: Rat) (n :: Rat) where
```
ChShersh: I discovered recently that operators can be written in infix form in definition, like this:

```haskell
type family (m :: Rat) ==% (n :: Rat) :: Rat where
```

(didn't try to compile this, so maybe it doesn't work actually).

Also, if you feel brave enough, you can try to make `==` operator overloaded just like @int-index did it for `*` and `/`. But I'm not sure it's possible with `==` because we need slightly different thing...

src/Time/Units.hs
```diff
@@ -132,6 +135,27 @@ type AllTimes =
...
+
+-- Drops wile not the required time unit in 'AllTimes'.
+type family DropWhileNot (a :: Rat) (c :: [Rat]) where
+    DropWhileNot x '[] = '[]
+    DropWhileNot x (u ': units) = If (u ==% x) (u ': units) (DropWhileNot x units)
+
+-- Takes while not equal to the provided bound.
+type family TakeWhileNot (a :: Rat) (c :: [Rat]) where
+    TakeWhileNot x '[] = '[]
+    TakeWhileNot x (u ': units)  = If (u ==% x) '[u] (u ': TakeWhileNot x units)
```
ChShersh: Looks cool :+1: I like this solution! I had in mind something much less elegant... So I really like the current one :slightly_smiling_face: Though here is redundant space before `=`.

src/Time/Units.hs
```diff
@@ -132,6 +135,27 @@ type AllTimes =
...
+{- | Creates the list of time units in descending order by provided
+the highest and the lowest bound of the desired list.
+
+__Usage example:__
+
+>>> seriesF @(Hour ... Second) $ hour 3 +: minute 5 +:sec 3 +: ms 123
+"3h5m3s"
+
+-}
+type family (...) (a :: Rat) (b :: Rat) where
```
ChShersh: I think it's better to give more meaningful names to types and variables. Like `from` and `to`.

src/Time/Units.hs
```diff
@@ -10,6 +10,7 @@
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE TypeApplications           #-}
 {-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE TypeInType                 #-}
```
ChShersh: Not sure, but I think `TypeInType` is redundant. Actually, `TypeInType` is almost never needed. This extension is rather complex and it allows to do some crazy things. We don't do such things in O'Clock :) Unfortunately, compiler is not so good at suggesting required extension...

src/Time/Bool.hs
int-index: There is already an overloaded type family for this is base:

https://hackage.haskell.org/package/base-4.10.1.0/docs/Data-Type-Equality.html#t:-61--61-

We need something like this to do overloading:

```
type instance m == n = m ==% n
```

Here we rely on the fact that `==%` has a fixed kind for the inputs (it is `Rat`), so the compiler will do kind-based dispatch.

ChShersh: @vrom911 tried something like this. Turned out that `==` is closed type family. The error is the following:

```haskell
Illegal instance for closed family ‘==’
• In the type instance declaration for ‘==’
    |
148 | type instance (a :% b) == (c :% d) = (a == c) And (b == d)
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

int-index: Not true, I just checked in GHCi, I can define instances of `(==)`.

int-index: I see, they changed it in 8.4

src/Time/Bool.hs
```diff
@@ -1,39 +1,38 @@
...
 -- | Type-level 'and'.
-type family And (a :: Bool) (b :: Bool) where
-  And 'False _ = 'False
-  And _      x = x
+type family (a :: Bool) && (b :: Bool) :: Bool where
+    'False && _ = 'False
+    _      && x = x
+
+-- | The result kind of overloaded equality.
+type family EqK (t :: Type) :: Type
```
int-index: This is not necessary, the result of `Eq` is always `Bool`.

src/Time/Bool.hs
```diff
@@ -1,39 +1,38 @@
...
+type instance EqK Rat = Bool
+
+-- | Overloaded equality.
+type family (a :: t) == (b :: t) :: EqK t
 
--- | A type family to compute equality of 'Rat's.
-type family (==%) (m :: Rat) (n :: Rat) where
-    (a :% b) ==% (c :% d) = (a == c) `And` (b == d)
+type instance (a :: Nat) == (b :: Nat) = a Equality.== b
+type instance (a :% b)   == (c :% d)   = (a == c) && (b == d)
```
int-index: Wait a second, so we're using structural equality here? Can you check please, perhaps `==` from `base` works out of the box?

vrom911: Ooh, sorry, you're totally right! We don't need to do anything, everything is already working..

src/Time/Bool.hs
```diff
@@ -1,39 +1,18 @@
...
+       ( If
+       , type (&&)
        ) where
 
-import Data.Type.Equality (type (==))
-
-import Time.Rational (type (:%), Rat)
-
 -- | Type-level 'if'.
 type family If (x :: Bool) a b where
```
ChShersh: Sorry, I probably need to go to doctor because I try to avoid `TypeInType` extension at any cost... I think you can replace `TypeInType` with `PolyKinds` and write if as `If (x :: Bool) (a :: k) (b :: k) :: k`. I guess this should work...

src/Time/Units.hs
```diff
@@ -144,18 +142,18 @@ __Usage example:__
 "3h5m3s"
 
 -}
-type family (...) (a :: Rat) (b :: Rat) where
+type family (from :: Rat) ... (to :: Rat) where
     x ... y = TakeWhileNot y (DropWhileNot x AllTimes)
```
ChShersh: Maybe variable names here can be changed from `x` and `y` to `from` and `to` as well.

src/Time/Units.hs
```diff
@@ -144,18 +142,18 @@ __Usage example:__
...
     DropWhileNot x '[] = '[]
-    DropWhileNot x (u ': units) = If (u ==% x) (u ': units) (DropWhileNot x units)
+    DropWhileNot x (u ': units) = If (u == x) (u ': units) (DropWhileNot x units)
 
 -- Takes while not equal to the provided bound.
-type family TakeWhileNot (a :: Rat) (c :: [Rat]) where
+type family TakeWhileNot (to :: Rat) (units :: [Rat]) where
     TakeWhileNot x '[] = '[]
-    TakeWhileNot x (u ': units)  = If (u ==% x) '[u] (u ': TakeWhileNot x units)
+    TakeWhileNot x (u ': units) = If (u == x) '[u] (u ': TakeWhileNot x units)
```
ChShersh: This is magic to me...  :sparkles:  I wasn't aware that `==` just works out of the box :ballot_box:

src/Time/Units.hs
```diff
@@ -144,18 +142,18 @@ __Usage example:__
 "3h5m3s"
 
 -}
-type family (...) (a :: Rat) (b :: Rat) where
+type family (from :: Rat) ... (to :: Rat) where
     x ... y = TakeWhileNot y (DropWhileNot x AllTimes)
 
 -- Drops wile not the required time unit in 'AllTimes'.
-type family DropWhileNot (a :: Rat) (c :: [Rat]) where
+type family DropWhileNot (from :: Rat) (units :: [Rat]) where
```
ChShersh: Could you please write result kinds of thess type families as well? I personally find it more useful. There's some type-level magic is going on here so I would like to have kind signatures for both arguments and result be explicit... With kinds specified explicitly it's easier to understand what's going on (heyeayayeayyey)

src/Time/Bool.hs
ChShersh: After discussion with @int-index I changed my mind. So you can keep `-XTypeInType`. Well, probably no harm in it.

vrom911:
Okay, I designed it so that when you write `[Hour ... Second]` and `[Second ... Hour]` it would behave the same way. Tell me if that is not the best choice..

ChShersh:
@vrom911 Nice! I like this solution. Probably, most often mistake will be to write `Microsecond ... Millisecond` or something like that. Current interface is good to me! I was thinking about type error in case arguments are not ordered. But current behavior is much more convenient. I think you can squash, I will review it once more and most likely approve.

int-index:
I'd prefer a type error when lower/upper bound are not in order.

ChShersh:
@int-index Could you explain why Type Error is better than arbitrary order?

int-index:
> Could you explain why Type Error is better than arbitrary order?

Because I believe that source code should reflect author's intention.

Let's consider two cases:

* person (A) is unware of the flipped semantics
* person (B) is aware of the flipped semantics

If formatting with `[Hour...Second]` corresponds to `1h5m35s`, then the intuitive semantics for person (A) is "write time components in specified order". Then, person (A) would expect `[Second...Hour]` to provide formatting like `35s5m1h`, which it doesn't, and it's unexpected and frustrating.

Now consider person (B). Why would he prefer to write `[Second...Hour]` over `[Hour...Second]`? Now he has an arbitrary decision to make each time he writes a time range. One option is better because its order corresponds to the resulting formatting, so the other option now seems redundant, and it becomes one more bit of arcane knowledge of how to do things right (a similar case is `return` vs `pure`, `fmap` vs `liftM`, etc, often it's hard to decide which one to use).

ChShersh:
@int-index Ok, sees legit and convincing.

PR #67, opened 2018-01-30
[#62] Implement several convenient utils

### LINKED ISSUE #62: Implement several convenient utils
Currently it's a little bit awkward to use `o-clock` in several places... This issue is devoted to collect such things:

- [x] Add `.ghci` to repository. It's not convenient to enable language extensions and add required imports manually every time when entering `cabal new-repl`.
- [x] Create time from rational time more conveniently. If you want _2/7 of day_ you can use one of the following ways (I don't find any of these ways convenient):
    1. `Time @Day (2 % 7)`
    2. `timeMul (1 % 7) (day 2)`
    3. `fromRational @(Time Day) (2 % 7)`
- [x] Subtract time of different units more conveniently. We have `+:` operator for `sec 3 +: minute 3`. I want something similar but for subtraction (see example in #60). Unfortunately, `-:` operator is very popular on Hackage. So we need another convention for arithmetic operators.

vrom911:
Resolves #62 

I decided not to add new functions to create time from rationals, so our helpers are now working with `RatioNat`. So you can use it like `sec (2 % 3)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The change log is available [on GitHub][2].
 0.0.1
 =====
 
+* [#62](https://github.com/serokell/o-clock/issues/62):
+  Add `.ghci` file. Make time creation helpers work with
+  `RatioNat` instead of `Naturals`.
+  Rename `+:` to `+:+` add `-:-`.
 * [#46](https://github.com/serokell/o-clock/issues/46):
   Introduce `...` type to create custom time unit lists in
   provided bounds.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The change log is available [on GitHub][2].
 0.0.1
 =====
 
+* [#62](https://github.com/serokell/o-clock/issues/62):
+  Add `.ghci` file. Make time creation helpers work with
+  `RatioNat` instead of `Natirals`.
```
ChShersh: Typo. Should be `Naturals` here.

PR #68, opened 2018-02-01
[#56] Add doctest support

### LINKED ISSUE #56: Add 'doctest' to 'o-clock'
Probably this will require some changes in documentation. But I still think it's a good idea to have tests with `doctest`. After adding such tests, we can remove duplicating tests from our test suite.

vrom911:
Resolves #56

### DIFF
--- a/src/Time/Formatting.hs
+++ b/src/Time/Formatting.hs
@@ -30,13 +30,15 @@ __Examples__
 The received list should be in descending order. It would be verified at compile-time.
 Example of the error from @ghci@:
 
+#if ( __GLASGOW_HASKELL__ >= 804 )
 >>> seriesF @'[Millisecond, Second] (minute 42)
-
-<interactive>:10:2: error:
+...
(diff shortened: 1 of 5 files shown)


src/Time/Formatting.hs
```diff
@@ -30,13 +30,15 @@ __Examples__
 The received list should be in descending order. It would be verified at compile-time.
 Example of the error from @ghci@:
 
->>> seriesF @'[Millisecond, Second] (minute 42)
+@
+__>>> seriesF @'[Millisecond, Second] (minute 42)__
```
ChShersh: Could you implement this as `doctest` test as well? Just see how I did here in `universum`:

* https://github.com/serokell/universum/blob/68727edc7c1c478a5de09677b2ab0b1e846f7615/src/Universum/Container/Class.hs#L604

You can use `...` to skip uninteresting parts. I think we should test this...

PR #73, opened 2018-02-04
[#72] Enhance TimeStamp, refactor

### LINKED ISSUE #72: Refactor and enhace time arithmetics
I think that everything related to arithmetic should be moved into `TimeStamp` module (like operators `-:-`, `+:+` etc.). Also, I want:

- [ ] `*:*` as alias to `timeMul`. `3 *:* sec 500` looks prettier than `timeMul 3 (sec 500)`
- [ ] `/:/` as alias to `timeDiv` (yey, more operators :harold:)
- [ ] Maybe we should have version of `-` operator for `Time` which returns `Ordering`, like for `TimeStamp`?
- [ ] `timeAdd` function should have type `timeAdd :: Time unit -> TimeStamp -> TimeStamp`

Maybe something else...

vrom911:
Resolves #72

### DIFF
--- a/src/Time/TimeStamp.hs
+++ b/src/Time/TimeStamp.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE DataKinds                  #-}
 {-# LANGUAGE ExplicitForAll             #-}
+{-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 -- | This module introduces 'TimeStamp' data type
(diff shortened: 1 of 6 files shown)


src/Time/TimeStamp.hs
```diff
@@ -10,19 +11,38 @@ module Time.TimeStamp
...
 
--- | Returns the result of comparison of two 'Timestamp's and
--- the 'Time' of that difference of given time unit.
+{- | Returns the result of comparison of two 'Timestamp's and
+the 'Time' of that difference of given time unit.
+
+>>> timeDiff @Second (TimeStamp 4) (TimeStamp 2)
+(GT,2s)
+
+>>> timeDiff @Second (TimeStamp 2) (TimeStamp 4)
```
ChShersh: Would be also good to add example with different time units, like, `Minute` or `Hour`

src/Time/TimeStamp.hs
```diff
@@ -49,9 +74,75 @@ timeMul :: forall (unit :: Rat) . KnownRat unit
         -> Time unit
 timeMul n (Time t) = Time (n * t)
 
+{- | Operator version of 'timeMul'.
+
+>>> 3 *:* sec 5
```
ChShersh: I think you should add `infixl` or `infixr` for this operator with some priority so something like this would be possible to write:

```haskell
ghci> 2 *:* 3 *:* sec 5
ghci> 3 *:* 5 *:* 7 :: Time Second
```

Not sure it's possible but looks possible...

src/Time/TimeStamp.hs
```diff
@@ -49,9 +74,75 @@ timeMul :: forall (unit :: Rat) . KnownRat unit
...
+      -> Time unitResult
+      -> (Ordering, Time unitResult)
+t1 -%- (Time t2Rat) =
+    let (Time t1Rat) = toUnit @unitResult t1
+        order = compare t1Rat t2Rat
+        newTime = Time $ case order of
+            LT -> t2Rat - t1Rat
+            GT -> t1Rat - t2Rat
+            EQ -> 0
+    in (order, newTime)
```
ChShersh: I think you can avoid code duplication with `timeDiff` function if you implement function which takes two rational numbers and subtract them with returning `Ordering`.

Issue #60, opened 2018-01-22
Make 'seriesF' implementation more precise


ChShersh:
It feels awkward to me that `seriesF` formatting is not precise enough:

```haskell
ghci> seriesF @'[Hour, Minute, Second] (Time @Day (2 % 7))
"6h51m25s"
ghci> seriesF @'[Hour, Minute, Second, Millisecond] (Time @Day (2 % 7))
"6h51m25s714ms"
```

Would be good to show fractional as a last argument in case time is not precise.

Like this:

```haskell
ghci> seriesF @'[Hour, Minute, Second] (Time @Day (2 % 7))
"6h51m180/7s"
```

Tested with:

```haskell
ghci> let t = Time @Day (2 % 7)
ghci> toUnit @Second (toUnit (toUnit t - hour 6) - minute 51)
180/7s
```

PR #75, opened 2018-02-04
Show/Read instances for Time now use mixed fractions

### LINKED ISSUE #61: Think about convenient way to show fractional time
Currently fractional time is shown as normalized fractional, like this:

```haskell
ghci> show $ Time @Second (180 % 7)
"180/7s"
```

This might be not so convenient... (see usage example in #60)

I think something like this would be more useful:

```haskell
ghci> show $ Time @Second (180 % 7)
"25+5/7s"
```

Not sure about format. Exactly this looks a little bit ugly to me... But maybe it's okay.

adituv:
The Show and Read instances for Time now support mixed fractions:
`show` always outputs a mixed fraction, and `read` can read either
an improper fraction (the previous functionality) or a mixed
fraction.

The Show instance now has `showsPrec` explicitly implemented
instead of `show` as precedence could potentially matter with the
'+' in the middle.  Parentheses are added when the surrounding
precedence is greater than 6, the precedence of '+'.  For example,
3 * (7+2/5s).  This should not affect `seriesF` as that uses show,
and so a precdence of 0.

Resolves #61

### DIFF
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -68,7 +68,7 @@ import GHC.Prim (coerce)
 import GHC.Read (Read (readPrec))
 import GHC.Real (denominator, numerator, (%))
 import GHC.TypeLits (KnownSymbol, Symbol, symbolVal)
-import Text.ParserCombinators.ReadP (ReadP, char, munch1, option, pfail)
+import Text.ParserCombinators.ReadP (ReadP, char, munch1, option, pfail, (+++))
 import Text.ParserCombinators.ReadPrec (ReadPrec, lift)
 
 #if ( __GLASGOW_HASKELL__ >= 804 )
@@ -145,11 +145,25 @@ unitNameVal :: forall (unit :: Rat) . (KnownUnitName unit) => String
 unitNameVal = symbolVal (Proxy @(UnitName unit))
 
 instance KnownUnitName unit => Show (Time unit) where
-    show (Time rat) = let numeratorStr   = show (numerator rat)
-                          denominatorStr = case denominator rat of
-                                                1 -> ""
-                                                n -> '/' : show n
-                      in numeratorStr ++ denominatorStr ++ unitNameVal @unit
+    showsPrec p (Time t) = showParen (p > 6)
+                              $ showsMixed t
+                              . showString (unitNameVal @unit)
+      where
+        showsMixed 0 = showString "0"
+        showsMixed rat =
+          let (n,d) = (numerator rat, denominator rat)
+              (q,r) = n `quotRem` d
+              op = if q == 0 || r == 0 then "" else "+"
+              quotStr = if q == 0
+                          then id -- NB id === showString ""
+                          else shows q
+              remStr = if r == 0
+                         then id
+                         else shows r
+                            . showString "/"
+                            . shows d
+          in
+              quotStr . showString op . remStr 
 
 instance KnownUnitName unit => Read (Time unit) where
     readPrec :: ReadPrec (Time unit)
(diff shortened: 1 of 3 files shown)


adituv:
Due to #60 (and because all the Show tests are implicit in seriesF
tests) I haven't implemented tests for the Show instance explicitly.
Let me know if you would like me to do that too, or whether you
would prefer to wait on #60.

I tried to look at #60 but I don't understand the 8.0.4+ type logic
enough to handle it right now

test/Test/Time/Units.hs
```diff
@@ -7,14 +7,17 @@ module Test.Time.Units
        ( unitsTestTree
        ) where
 
-import Control.Exception (evaluate)
-import GHC.Real (Ratio ((:%)))
-import Test.Tasty (TestTree)
-import Test.Tasty.Hspec (Spec, anyException, describe, it, shouldBe, shouldThrow, testSpec)
+import           Control.Exception (evaluate)
```
ChShersh: I pushed our company's `.stylish-haskell.yaml` to `master` branch. Could you please use this `.stylish-haskell.yaml` config for formatting  imports? All you need is just rebase on `master` in your fork/branch and then run `stylish-haskell` on this file again.

test/Test/Time/Units.hs
```diff
@@ -45,6 +48,14 @@ spec_Units = do
...
         it "fails when '-4s' is expected as seconds" $
             evaluate (read @(Time Second) "-4s") `shouldThrow` anyException
+        it "parses '25+5/7s' as 180/7 seconds" $
+            read @(Time Second) "25+5/7s" `shouldBe` Time (180 :% 7)
+        it "fails when '3+2s' is expected as seconds" $
+            evaluate (read @(Time Second) "3+2s") `shouldThrow` anyException
+        it "fails when '+3s' is expected as seconds" $
+            evaluate (read @(Time Second) "+3s") `shouldThrow` anyException
+        it "fails when '/3s' is expected as seconds" $
+            evaluate (read @(Time Second) "/3s") `shouldThrow` anyException
```
ChShersh: Unit tests are great! It's nice to see which cases should fail.

src/Time/Units.hs
```diff
@@ -162,11 +176,19 @@ instance KnownUnitName unit => Read (Time unit) where
...
             let naturalP = read <$> munch1 isDigit
-            n <- naturalP
-            m <- option 1 (char '/' *> naturalP)
+            -- If a '+' is parsed as part of a mixed fraction, the other parts
+            -- are no longer optional.  This separation is required to prevent
+            -- e.g. "3+2" successfully parsing.
+            let fullMixedExpr = (,,) <$> (naturalP <* char '+')
+                                     <*> (naturalP <* char '/')
+                                     <*> naturalP
+            let improperExpr = (,,) 0 <$> naturalP
```
ChShersh: Using `-XTupleSections` extension you can write this as:

```haskell
let improperExpr = (0,,) <$> naturalP
                         <*> option 1 (char '/' *> naturalP)
```

Just a minor suggestions 🙂

src/Time/Units.hs
```diff
@@ -162,11 +176,19 @@ instance KnownUnitName unit => Read (Time unit) where
         readP :: ReadP (Time unit)
         readP = do
             let naturalP = read <$> munch1 isDigit
-            n <- naturalP
-            m <- option 1 (char '/' *> naturalP)
+            -- If a '+' is parsed as part of a mixed fraction, the other parts
+            -- are no longer optional.  This separation is required to prevent
+            -- e.g. "3+2" successfully parsing.
```
vrom911: I still can write `3+2/1s` and it will be parsed as `5s` :stuck_out_tongue_winking_eye:. So I think at this moment we need to decide whether we should also forbid using `1` in the denominator, because it's kinda similar cases for me..
What are your thoughts @ChShersh

ChShersh: I don't know really... Tough decisions 🤔

ChShersh:
Oh, one more thing: could you please update changelog as well?

adituv:
Sorry, I seem to have messed up something on git!  I'm fixing it now.

src/Time/Units.hs
adituv: Hmm, true, and I don't enforce correct separation of mixed fractions either.  This could be fixed by doing that (so that the remainder part has to have numerator < denominator)

adituv:
Git problems resolved.  I am not sure on the protocol for when responding to requested changes though: should I squash the commits?  For now I've left them separated.

ChShersh:
@adituv Sorry, we don't have contributing guide at this project at the moment 😞 So it's up to you. It makes sense to me to squash commits after code review (to make code review easier and resulting history cleaner). But I don't insist on this for volunteer contributors 🙂

src/Time/Units.hs
vrom911: Yes, true.. 
Let's keep it as it is now for the moment :slightly_smiling_face:

ChShersh:
@adituv Thanks for your amazing contribution to our project! 🎉

PR #77, opened 2018-02-05
[#76] Make TimeStamp work with Seconds internally

### LINKED ISSUE #76: Reimplement 'TimeStamp' internal representation
It was proposed to store `TimeStamp` as seconds internally and make it an implementation detail.

So in future if something like this is written:

```haskell
ts <- getCurrentTime
threadDelay (2 :: Minute)
ts' <- getCurrentTime
print (timeDiff @Minute ts ts')
print (timeDiff @Hour ts ts')
```

The following will be printed: approximately `2m` followed by approximately `1/30h`.

Also, let's remove numeric instances for `TimeStamp`.

vrom911:
Resolves #76

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,12 @@ Change log
 o'clock uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
-0.0.1
+0.1.0
 =====
 
+* [#76](https://github.com/serokell/o-clock/issues/76):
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The change log is available [on GitHub][2].
 0.0.1
```
ChShersh: Could you also change version here to `0.1.0`? I guess we will release new version of `o-clock` in some nearest future... And it's better to have major version bumped up.

PR #87, opened 2018-02-10
[#63] Add SeriesP class

### LINKED ISSUE #63: Implement parser for 'seriesF' result
We have nice formatting with `seriesF` function. Would be nice to have parser for this format. I can think of the following API:

```haskell
ghci> seriesP @Minute "2d18h40m"
4000m
```

vrom911:
Resolves #63

### DIFF
--- a/src/Time/Series.hs
+++ b/src/Time/Series.hs
@@ -8,64 +8,36 @@
 {-# LANGUAGE TypeOperators        #-}
 {-# LANGUAGE UndecidableInstances #-}
 
-{- | This module introduces function to format time in desired way.
+-- | This module introduces function to format and parse time in desired way.
 
-__Examples__
-
->>> seriesF @'[Day, Hour, Minute, Second] (minute 4000)
-"2d18h40m"
-
->>> seriesF @'[Day, Minute, Second] (minute 4000)
-"2d1120m"
-
->>> seriesF @'[Hour, Minute, Second] (sec 3601)
-"1h1s"
-
->>>  seriesF @'[Hour, Second, Millisecond] (Time @Minute $ 3 % 2)
-"90s"
-
->>> seriesF @'[Hour, Second] (minute 0)
-"0h"
-
->>> seriesF @'[Hour, Minute, Second] (Time @Day (2 % 7))
-"6h51m25+5/7s"
-
-The received list should be in descending order. It would be verified at compile-time.
-Example of the error from @ghci@:
-
-#if ( __GLASGOW_HASKELL__ >= 804 )
->>> seriesF @'[Millisecond, Second] (minute 42)
-...
-    • List of units should be in descending order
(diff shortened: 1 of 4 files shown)


src/Time/Series.hs
```diff
@@ -174,3 +181,79 @@ instance ( KnownRatName unit
...
+    seriesP :: forall (someUnit :: Rat) . KnownRatName someUnit
+            => String -> Maybe (Time someUnit)
+
+instance SeriesP '[] where
+    seriesP _ = Nothing
+
+instance (KnownRatName unit) => SeriesP '[unit] where
+    seriesP :: forall (someUnit :: Rat) . KnownRatName someUnit
+            => String -> Maybe (Time someUnit)
+    seriesP ""  = Just $ Time 0
```
ChShersh: I think this case (parsing empty string) should be in documentation. It's a hard problem, how we want to handle empty string. But this is not important at the moment. What's important is not to have undocumented behavior.

src/Time/Series.hs
```diff
@@ -174,3 +181,79 @@ instance ( KnownRatName unit
...
+>>> seriesP @'[Day, Minute, Second] @Minute "2d1120m"
+Just (4000m)
+
+>>> seriesP @'[Hour, Minute, Second] @Second "1h1s"
+Just (3601s)
+
+>>> seriesP @'[Hour, Second, Millisecond] @Minute "90s"
+Just (1+1/2m)
+
+>>> seriesP @'[Hour, Second] @Second "11ns"
```
ChShersh: Could you also add examples with something like `seriesP @'[Hour, Minute]` where you have `1+1/2h` and `1+1/2m` and `1h1+1/2m`?

src/Time/Series.hs
```diff
@@ -174,3 +181,79 @@ instance ( KnownRatName unit
...
+         => SeriesP (unit ': nextUnit ': units :: [Rat]) where
+    seriesP :: forall (someUnit :: Rat) . KnownRatName someUnit
+            => String -> Maybe (Time someUnit)
+    seriesP ""  = Just $ Time 0
+    seriesP str = let (num, rest)  = span isDigit str
+                      (u, nextStr) = span isLetter rest
+                      maybeT = readMaybeTime @unit $ num ++ u
+                  in case maybeT of
+                         Nothing -> seriesP @(nextUnit ': units) str
+                         Just _  -> liftA2 (+) maybeT (seriesP @(nextUnit ': units) nextStr)
```
ChShersh: Personally, I don't like ignoring result of `Just` here. Here you perform manual unwrapping of `maybeT` and then you pass `maybeT` to `liftA2` function which will do this unwrapping again. It looks better to me to use `fmap` here, like

```haskell
Just t -> (t+) <$> seriesP ...
```

PR #89, opened 2018-02-11
[#71] Add toNum function

### LINKED ISSUE #71: Implement 'toNum' function
Some libraries still use `Integer` to represent time. It would be nice to have some `toNum` function so instead of 

```haskell
mskDifference :: POSIXTime
mskDifference = 10800 -- 3 hours
```

We could write in more safe way:

```haskell
mskDifference :: POSIXTime
mskDifference = toNum @Second $ hour 3
```

Current possible solution:

```haskell
mskDifference :: POSIXTime
mskDifference = floor $ toUnit @Second $ hour 3
```

Dunno is it worth to have such conversion function...

vrom911:
Resolves #71

### DIFF
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -37,6 +37,7 @@ module Time.Units
         -- ** Creation helpers
        , time
        , floorUnit
+       , toNum
 
        , sec
        , ms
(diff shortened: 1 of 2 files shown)


src/Time/Units.hs
```diff
@@ -312,6 +313,22 @@ fortnight = time
 floorUnit :: forall (unit :: Rat) . Time unit -> Time unit
 floorUnit = time . fromIntegral @Natural . floor
 
+{- | Convert time to the 'Num' in given units.
```
ChShersh: I think it's always useful to add use cases in documentation. For example, would be good to write, that instead of writing `foo :: POSIXTime foo = 10800` one can write `foo = toNum @Second $ hour 3`. So people could learn how and why use some functions. They are exist on purporse (functions, not users).

PR #96, opened 2018-02-26
[#95] Add Semigroup and Monoid instances for Time

### LINKED ISSUE #95: Add 'Semigroup' and 'Monoid' instances for 'Time'
Since `Time` data type describes _time interval_ or _time duration_ it makes total sense to implement such instances. Especially when multiplication of times doesn't make sense. So there's only single correct way to implement such instances.

vrom911:
Resolves #95

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,9 +24,7 @@ TAGS
 *#
 
 cabal.project.local
-.ghc.environment.x86_64-linux-8.4.0.20171222
-.ghc.environment.x86_64-linux-8.4.0.20180118
-.ghc.environment.x86_64-linux-8.4.0.20180204
+.ghc.environment.*
 
 # IDE/support
 .idea/
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -62,7 +62,9 @@ import Control.Applicative ((*>))
 import Control.Monad (unless)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Char (isDigit, isLetter)
+import Data.Foldable (foldl')
 import Data.Proxy (Proxy (..))
+import Data.Semigroup (Semigroup (..))
 import GHC.Generics (Generic)
 import GHC.Natural (Natural)
 import GHC.Prim (coerce)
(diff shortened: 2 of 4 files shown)


src/Time/Units.hs
```diff
@@ -120,6 +124,13 @@ type Fortnight   = 1209600 :% 1
 newtype Time (rat :: Rat) = Time { unTime :: RatioNat }
     deriving (Eq, Ord, Enum, Real, RealFrac, Generic)
 
+instance Semigroup (Time (rat :: Rat)) where
+    (<>) = (+)
+
+instance Monoid (Time (rat :: Rat)) where
+    mempty  = Time 0
+    mappend = (<>)
```
ChShersh: We need to implement `sconcat` (from `Semigroup`) and `mconcat` (from `Monoid) using `foldl'` function because default implementations of these functions use `foldr` which makes them extremely inefficient.

src/Time/Units.hs
```diff
@@ -120,6 +122,21 @@ type Fortnight   = 1209600 :% 1
 newtype Time (rat :: Rat) = Time { unTime :: RatioNat }
     deriving (Eq, Ord, Enum, Real, RealFrac, Generic)
 
+-- | Addition is associative binary operation for 'Semigroup' of 'Time'.
+instance Semigroup (Time (rat :: Rat)) where
+    (<>) = (+)
+    {-# INLINE (<>) #-}
+    sconcat = foldl' (<>) mempty
+    {-# INLINE sconcat #-}
```
ChShersh: OMG, I'm stupid. `stimes` also can be implemented efficiently using multiplication. Checkout `Semigroup` instance for `Sum`:

* http://hackage.haskell.org/package/base-4.10.1.0/docs/src/Data.Semigroup.html#line-243

.gitignore
```diff
@@ -27,6 +27,7 @@ cabal.project.local
 .ghc.environment.x86_64-linux-8.4.0.20171222
 .ghc.environment.x86_64-linux-8.4.0.20180118
 .ghc.environment.x86_64-linux-8.4.0.20180204
+.ghc.environment.x86_64-linux-8.4.0.20180224
```
int-index: You can do `.ghc.environment.*`

PR #101, opened 2018-03-07
[#100] Add instances

### LINKED ISSUE #100: Add more instances to 'o-clock' library
I see that in `safe-money` package they choose the following approach for having `aeson` instances and other: they introduce flags in `.cabal` file and depend on each library conditionally. I think we can do the same thing:

* https://github.com/k0001/safe-money/blob/master/safe-money.cabal

Not sure we want all instances, but I think it's worth to have the following ones:

- [x] `Hashable` (`hashable`)
- [x] `NFData` (`deepseq`)
- [x] `ToJSON/FromJSON/ToJSONKey/FromJSON/Key` (`aeson`, also maybe with one extra `newtype` for additional behavior with `mcs` suffixes and without)
- [x] `serialise` (I saw this library was used in different projects)

vrom911:
Resolves #100

### DIFF
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -82,6 +82,10 @@ import Data.Hashable (Hashable)
 import Control.DeepSeq (NFData)
 #endif
 
+#ifdef HAS_serialise
+import Codec.Serialise (Serialise (..))
+#endif
+
 #if ( __GLASGOW_HASKELL__ >= 804 )
 import Time.Rational (type (*), type (/))
 #endif
@@ -153,6 +157,10 @@ instance Hashable (Time (rat :: Rat))
 
 #ifdef HAS_deepseq
 instance NFData (Time (rat :: Rat))
+#endif
+
+#ifdef HAS_serialise
+instance Serialise (Time (rat :: Rat))
 #endif
 
 -- | Type family for prettier 'show' of time units.
--- a/src/Time/Units.hs
+++ b/src/Time/Units.hs
@@ -74,6 +74,10 @@ import GHC.TypeLits (KnownSymbol, Symbol, symbolVal)
 import Text.ParserCombinators.ReadP (ReadP, char, munch1, option, pfail, (+++))
 import Text.ParserCombinators.ReadPrec (ReadPrec, lift)
 
+#ifdef HAS_hashable
+import Data.Hashable (Hashable)
+#endif
+
 #if ( __GLASGOW_HASKELL__ >= 804 )
 import Time.Rational (type (*), type (/))
 #endif
@@ -138,6 +142,10 @@ instance Monoid (Time (rat :: Rat)) where
     {-# INLINE mappend #-}
     mconcat = foldl' (<>) mempty
     {-# INLINE mconcat #-}
+
+#ifdef HAS_hashable
+instance Hashable (Time (rat :: Rat))
+#endif
 
 -- | Type family for prettier 'show' of time units.
 type family UnitName (unit :: Rat) :: Symbol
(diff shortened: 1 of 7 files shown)


src/Time/Units.hs
```diff
@@ -163,6 +169,15 @@ instance NFData (Time (rat :: Rat))
 instance Serialise (Time (rat :: Rat))
 #endif
 
+#ifdef HAS_aeson
+instance (KnownUnitName unit) => ToJSON (Time (unit :: Rat)) where
+    toJSON = toJSON . show
+
+instance (KnownUnitName unit) => FromJSON (Time (unit :: Rat)) where
+    parseJSON (String t) = pure $ read @(Time unit) $ Text.unpack t
```
ChShersh: It's better to use `readMaybe` here and report error with `fail` function. Otherwise `read` will throw pure exception which we can't handle.

Also, instead of manual pattern-matching on `String` it's better to use [`withText`](https://hackage.haskell.org/package/aeson-1.2.4.0/docs/Data-Aeson.html#v:withText) function.

src/Time/Units.hs
```diff
@@ -163,6 +169,15 @@ instance NFData (Time (rat :: Rat))
 instance Serialise (Time (rat :: Rat))
 #endif
 
+#ifdef HAS_aeson
+instance (KnownUnitName unit) => ToJSON (Time (unit :: Rat)) where
+    toJSON = toJSON . show
```
ChShersh: I like the default behavior. But let's also implement `newtype NatTime unit` with JSON instances like in `yt-utilities`. I think its quite useful in many cases.

vrom911:
I'm sorry I don't understand why do we need to make tests on instances?

ChShersh:
@vrom911 As I see now, the only non-trivial instance we have is `ToJSON/FromJSON` instance. It's very simple but still non-trival. So I think tests if needed then needed only for this. Typical roundtrip tests like `fromJSON . toJSON ≡ id`. Though, this instance uses `show` and `read` and nothing more and we already have `read . show ≡ id` tests so may there's no much sense in having tests for `JSON` while we have simple instances.

src/Time/Units.hs
vrom911: But that's going to be not type safe instance then :thinking:  I mean `time: 5` can be minute, second, etc. and that could lead to some issues. The one that I've implemented uses `read` that's why it would work as I expect it to be..

ChShersh: @vrom911 Hmm, you're right. We don't want to motivate unsafe behavior. I guess if people work with unsafe API they should do some tricks to work carefully. Changing type of unit shouldn't go without problems. So, if thinking more about `NatTime` newtype it probably shouldn't exist.

vrom911: Okay :+1: Thanks for clarifications!

src/Time/Units.hs
```diff
@@ -163,6 +170,16 @@ instance NFData (Time (rat :: Rat))
...
+
+instance (KnownUnitName unit) => FromJSON (Time (unit :: Rat)) where
+    parseJSON = withText "time" $ \t -> fromMaybe
+        (fail $ "Can not parse Time. Expected unit: " ++ unitNameVal @unit)
+            $ pure <$> (readMaybe @(Time unit) $ Text.unpack t)
```
ChShersh: I think this can be written a little bit shorter and probably clearer with the help of `maybe` function. 

`fromMaybe x $ pure <$> t` is equivalent to `maybe x pure t`. Unfortunately, `base` package doesn't have `whenNothing` combinator which is exactly `maybe x pure t` :disappointed: 

You also can go with pointfree and write something like `maybe . readMaybe . Text.unpack` (but maybe this will be less clearer.

PR #104, opened 2018-04-02
[#103] Use latest nightly resoler for ghc-8.4.1

### LINKED ISSUE #103: Use Stackage nightly resolver for `GHC-8.4.1`
https://www.reddit.com/r/haskell/comments/842e2k/stackage_nightly20180313_with_ghc_841_is_out/

vrom911:
Resolves #103


stack.yaml
```diff
@@ -1,98 +1,12 @@
...
-- time-locale-compat-0.1.1.3
-- unordered-containers-0.2.9.0
-- uuid-types-1.0.3
-
-- integer-logarithms-1.0.2.1
-- old-locale-1.0.0.7
-
-- happy-1.19.9
 
 allow-newer: true
```
ChShersh: Is `allow-newer` still needed? Also, do we need to patch `.travis.yml` or not? I don't remember...

I guess, CHANGELOG update is not required.. And after this PR is merged we can safely upload `1.0.0` version to Hackage!

vrom911: Well, I didn't try to remove it, but then `base` and a lot of other packages should be added in `stack.yaml`.. Should I try?

PR #107, opened 2018-06-21
[#106] Remove Num and related instances for Time

### LINKED ISSUE #106: Remove `Num` instance?
We already have had the conversation about `Num` instance for `Time` (#14 – issue, #13 – discussion in comments). Time multiplication doesn't make any sense so we agreed on having `error` on illegal actions from `Num` class just to have `fromInteger` that we needed for convenience. And, based on feedback, people don't agree that `error` is acceptable way, so it's better to get rid of any distrustful code we have. I would like to ask you @ChShersh @int-index  if you're also voting for removing these instances?

vrom911:
Resolves #106

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ module Main where
 #if ( __GLASGOW_HASKELL__ >= 804 )
 import Time (type (*))
 #endif
-import Time ((:%), (-:-), Time (Time), Hour, UnitName,floorUnit, hour, seriesF, toUnit)
+import Time ((:%), (-:-), Time, Hour, UnitName,floorUnit, hour, seriesF, toUnit)
 
 ```
 
@@ -121,8 +121,7 @@ Now let's implement main logic of our application. Our main function should take
 convert them to work weeks and work days and then show in human readable format.
 
 ```haskell
-calculateWork :: Time Hour  -- type synonym for 'Time HourUnit'
-              -> (Time WorkWeek, Time WorkDay)
+calculateWork :: Time Hour -> (Time WorkWeek, Time WorkDay)
 calculateWork workHours =
     let completeWeeks = floorUnit $ toUnit @WorkWeek workHours
         completeDays  = floorUnit $ toUnit @WorkDay  workHours -:- toUnit completeWeeks
@@ -140,7 +139,7 @@ So the similar result (but not rounded) can be gained with the usage of it. Chec
 ```haskell
 main :: IO ()
 main = do
-    putStrLn $ "The result:   " ++ formatHours (Time 140)
+    putStrLn $ "The result:   " ++ formatHours (hour 140)
     putStrLn $ "With seriesF: " ++ (seriesF @'[WorkWeek, WorkDay] $ hour 140)
 ```
 
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Hackage](https://img.shields.io/hackage/v/o-clock.svg)](https://hackage.haskell.org/package/o-clock)
 [![Build status](https://travis-ci.org/serokell/o-clock.svg?branch=master)](https://travis-ci.org/serokell/o-clock)
+[![Stackage](http://stackage.org/package/o-clock/badge/lts)](http://stackage.org/lts/package/o-clock)
 [![Stackage Nightly](http://stackage.org/package/o-clock/badge/nightly)](http://stackage.org/nightly/package/o-clock)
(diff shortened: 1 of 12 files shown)


src/Time/Series.hs
```diff
@@ -162,10 +164,15 @@ instance ( KnownRatName unit
...
                     timeStr = case flooredNewUnit of
-                                   0 -> ""
+                                   Time 0 -> ""
                                    _ -> show flooredNewUnit
-                    nextUnit = newUnit - flooredNewUnit
-                in if nextUnit == 0
+
+#if ( __GLASGOW_HASKELL__ >= 804 )
+                    nextUnit = withRuntimeDivRat @unit @unit $ newUnit -:- flooredNewUnit
+#else
```
ChShersh: Damn, I'm waiting for that day when we will drop support for GHC < 8.4.
Or when there appear good way to handle multiple versions.

test/Test/Time/Units.hs
```diff
@@ -54,20 +53,20 @@ spec_Units = do
         it "fails when '/3s' is expected as seconds" $
             evaluate (read @(Time Second) "/3s") `shouldThrow` anyException
         it "parses '14/2h' as 7 hours" $
-            read @(Time Hour) "14/2h" `shouldBe` 7
+            read "14/2h" `shouldBe` hour 7
         it "fails when '14/2h' expected as 7 seconds" $
             evaluate (read @(Time Second) "14/2h") `shouldThrow` anyException
         it "parses big number to big number" $
-            read @(Time Microsecond) ('1' : replicate 20 '0' ++ "mcs") `shouldBe` 100000000000000000000
```
ChShersh: I think this number can be written as `(10 ^ (20 :: Int))`

test/Test/Time/Timestamp.hs
```diff
@@ -22,18 +22,18 @@ spec_Timestamp :: Spec
...
+            timeDiff @Week (Timestamp 42) (Timestamp 42) `shouldBe` (EQ, Time 0)
         it "3 hours offset 7 is 10"  $
-            timeAdd @Hour 3 (Timestamp 7) `shouldBe` (Timestamp 10807)
+            timeAdd (hour 3) (Timestamp 7) `shouldBe` (Timestamp 10807)
         it "twice 21 mcs is 42 mcs"  $
-            timeMul @Microsecond 2 21 `shouldBe` 42
+            timeMul 2 (ms 21) `shouldBe` ms 42
         it "zero x 42 s is zero"  $
-            timeMul @Second 0 42 `shouldBe` 0
+            timeMul 0 (sec 42) `shouldBe` sec 0
```
ChShersh: I imagine this pain in fixing all these tests...

README.md
```diff
@@ -124,7 +125,7 @@ calculateWork :: Time Hour  -- type synonym for 'Time HourUnit'
               -> (Time WorkWeek, Time WorkDay)
 calculateWork workHours =
     let completeWeeks = floorUnit $ toUnit @WorkWeek workHours
-        completeDays  = floorUnit $ toUnit @WorkDay  workHours - toUnit completeWeeks
+        completeDays  = floorUnit $ toUnit @WorkDay  workHours -:- toUnit completeWeeks
```
ChShersh: I guess second `toUnit` is redundant. Could you try to remove it and run this executable to see result?

README.md
```diff
@@ -124,7 +125,7 @@ calculateWork :: Time Hour  -- type synonym for 'Time HourUnit'
               -> (Time WorkWeek, Time WorkDay)
```
ChShersh: Could you also delete comment about type synonym on line above? I forgot to delete it earlier...

README.md
```diff
@@ -139,7 +140,7 @@ So the similar result (but not rounded) can be gained with the usage of it. Chec
 ```haskell
 main :: IO ()
 main = do
-    putStrLn $ "The result:   " ++ formatHours 140
+    putStrLn $ "The result:   " ++ formatHours (Time 140)
```
ChShersh: I think it's slightly better for documentation to use `hour` function instead of `Time` constructor

o-clock.cabal
```diff
@@ -14,10 +14,10 @@ build-type:          Simple
 extra-doc-files:     CHANGELOG.md
```
ChShersh: Could you also please add `stability: stable` to these fields?

* https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-stability

I think it can be slightly better to show that library is stable than to not show it.

README.md
vrom911: Then the types won't match (we want `completeDays` to be `Time WorkDay`)

PR #115, opened 2021-01-10
Allow tasty-1.4

jkachmar:
Related to commercialhaskell/stackage#5795; I haven't tested this locally, I just figured it would be easy to make the change and let CI yell at me if something broke.


gromakovsky:
Thanks, I have ported this change to #117 because I also had to make another change to fix CI (#116). It's now available in https://hackage.haskell.org/package/o-clock-1.2.0.1

PR #119, opened 2021-03-18
Update doctest upper bound and some configuration

gromakovsky:
Simple maintenance PR that just updates some versions. I suppose we don't need a new release for it, we can just make a new revision.


gromakovsky:
Since nobody actively maintains `o-clock` and the changes here are fairly trivial, I think it's fine if I just merge it w/o approval as soon as CI passes.

Issue #124, opened 2021-08-30
Avoid accidental flooring when converting `Time unit` to numeric types


dcastro:
I found it unexpected that `toNum` floors the number before converting it to `Num a`. This is neither documented nor (IMO) something most users would expect to happen.

```hs
toNum @Second @Double (sec 1.4)
1.0
toNum @Second @Rational (sec 1.4)
1 % 1
```

This can lead to surprising results:

```hs
λ> toUnit @Second $ ms 999
999/1000s
λ> toNum @Second @Rational $ toUnit @Second $ ms 999
0 % 1
```

AFAICT, there is no way to convert `Time unit` to _an arbitrary_ `Num a => a` without flooring the ratio first.

So, I propose we deprecate `toNum` (and eventually delete it) and split it into `floorRat` (which already exists) and `toFractional`:

* add `toFractional` to support properly converting to Double, Float, etc:
    1. We could losslessly convert `Time unit` -> `Ratio Natural` -> `Ratio Integer`
    2. and then use `fromRational :: Fractional a => Ratio Integer -> a`
* to convert to integral types, we already have `floorRat`:
    * `floorRat` has the right constraint to prevent surprises like the ones I described above (`Integral` instead of `Num`) and a name that better describes what it does.

dcastro:
Re-opening this issue.

We've deprecated `toNum` in v1.3.0. Let's delete it in the next release.

PR #129, opened 2022-06-08
[#124] Add lossless conversion to Fractional values

breakerzirconia:
Problem: The function `toNum` transforms an object of the type 
`Time timeunit` to a value of the type that satisfies the `Num` 
constraint. However, the former object can contain a fractional value
inside, which then needs to get floored before being converted to a
`Num` value. This could lead to a possible loss of precision.

Solution: Create a separate function that converts the `Time` objects
losslessly to a value of the type that satisfies a stricter constraint:
`Fractional`. The lossless conversion can be done via `fromRational`.


src/Time/Units.hs
```diff
@@ -348,6 +349,28 @@ __Examples:__
...
+10800 % 1
+
+>>> toFractional @Hour @Rational $ hour 3
+3 % 1
+
+>>> toFractional @Day @Rational $ hour 3
+1 % 8
+
+>>> toFractional @Day @Double $ hour 3
+0.125
```
dcastro, line 368: Great examples! :clap:

src/Time/Units.hs
```diff
@@ -348,6 +349,28 @@ __Examples:__
...
+
+-}
+toFractional :: forall (unitTo :: Rat) r (unit :: Rat) . (KnownDivRat unit unitTo, Fractional r)
+             => Time unit -> r
+toFractional = fromRational . toRational . unTime . toUnit @unitTo
```
dcastro, line 373: I don't think `toFractional` should internally call `toUnit`.

This implementation assumes the input `Time unit` isn't _yet_ in the desired fractional unit.
But, if it is, then the call to `toUnit` is needless work, like in your second example:

```hs
>>> toFractional @Hour @Rational $ hour 3
3 % 1
```

IMO `toFractional` should do _only_ it what says on the tin: convert the `Time unit` to a fractional number.
If the user wants to convert the units first, then they can compose `toFractional` with `toUnit` themselves.

(PS: Yes, I know `toNum` does this, but IMO it shouldn't, and that's another reason I'm glad we're getting rid of it :sweat_smile: )

dcastro: Oh, after you make this change, please make sure the type variables are ordered according to #130

src/Time/Units.hs
```diff
@@ -44,6 +44,7 @@ module Time.Units
        , floorUnit
```
dcastro, line 44: Oh, and please add an "Unreleased" section to the changelog, and an entry under that explaining what changed and how to migrate.

src/Time/Units.hs
```diff
@@ -300,7 +301,7 @@ fortnight = time
 {-# INLINE fortnight #-}
 
 -- | Returns the greatest integer not greater than given 'Time'.
-floorRat :: forall (unit :: Rat) b . (Integral b) => Time unit -> b
+floorRat :: forall b (unit :: Rat) . (Integral b) => Time unit -> b
```
dcastro, line 304: Oh cool, you ended up doing #130 in this PR. Let's add this to the changelog as well then ^^

dcastro (approved):
Good stuff, feel free to squash the fixups and merge ^^ :+1:

PR #133, opened 2022-10-05
[Chore] Update upper bounds for `base`, `hedgehog` and `tasty-hedgehog`

Sereja313:
Problem: Newer versions of `base`, `hedgehog` and `tasty-hedgehog` are available, we need to add support for these versions. Also, we need to update stackage resolver to the latest LTS.

Solution: Update the upper bounds for these packages. Update stackage resolver to `lts-19.27`.

### DIFF
--- a/o-clock.cabal
+++ b/o-clock.cabal
@@ -23,8 +23,8 @@ extra-doc-files:     CHANGELOG.md
 tested-with:         GHC == 8.6.5
                    , GHC == 8.8.3
                    , GHC == 8.10.4
-                   , GHC == 9.0.1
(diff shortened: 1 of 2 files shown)


o-clock.cabal
```diff
@@ -32,7 +32,7 @@ source-repository head
 
 common common-options
   build-depends:
-      base >=4.12 && <4.17
+      base >=4.12 && <4.18
```
dcastro, line 35: I think it should be safe to set base's upper bound to `< 5`. IME, most packages do that.

